PR #1, opened 2019-12-10
Change update to draw

Otteri:
Update call doesn't work with Python 3.7.2, but draw works. Draw can be
also found from current matplotlib documentation (v3.1.1) whereas update
does not seem to exist for 'FigureCanvasTKAgg'.


wkirgsn:
Thanks for the hint.
We are working on a major update of the repo where this issue will be resolved among others. Stay tuned!

PR #22, opened 2020-04-24
correct coverage calculation

stheid:


wkirgsn:
Thanks for pointing this out!

PR #53, opened 2020-06-22
remove qt5 dependency from requirements after vis refactoring

wkirgsn:


wkirgsn:
Mere removal won't work. Travis says no

PR #59, opened 2020-06-27
Issue 21 rev

RohithCharanD:


examples/perm_dc_omega.py
```diff
@@ -1,29 +1,48 @@
...
+
+        Limitations: Physical limitations of the motor like Current,Speed
+
+        Converter : FourQuadrantConverter from converters.py
+
+"""
+
 if __name__ == '__main__':
+
+    """Discrete mode: It is the mode where the input command causes the controlled variable to change its
```
wallscheid, line 28: The explanations to both, the continuous as well as the discrete action mode are not correct:

Continuous mode: action is the average (normalized) voltage per time step which is assumed to be transferred to a PWM unit to generate the switching sequence (not part of the model, since we are working with averaged quantities). Hence, the action a continuous quantity.

Discrete mode: the action is the switching state  of the power converter i.e. a quantity from a discrete set of switching states which can be generated by the converter. Hence we directly generate the switching sequence by this control approach which default step time is therefore much smaller compared to the continuous time framework.

examples/perm_dc_omega.py
```diff
@@ -1,29 +1,48 @@
...
+Run this file from within the 'examples' folder:
+>> cd examples
+>> python perm_dc_omega.py
+
+Description:
+        Environment to control a Continuously controlled DC Permanently Excited Motor.
+
+        Controlled Quantity: 'omega'
+
+        Limitations: Physical limitations of the motor like Current,Speed
```
wallscheid, line 20: Please be more specific: we have an active state monitoring system which will terminate an episode whenever a motor state (such as current, speed,...) are violated (currently defined as box constraints on the states). Moreover, I'm missing the information that in case of using only a single loop speed control mode (as per default of this example), an increased number of motor current violations might be expected.

examples/perm_dc_omega.py
```diff
@@ -1,29 +1,48 @@
...
+    """
     env = gem.make(
-        'DcPermExDisc-v1',  # replace with 'DcPermExCont-v1' for continuous mode
-        # Pass an instance
-        visualization=MotorDashboard(plotted_variables=['omega', 'torque', 'i', 'u', 'u_sup'], visu_period=1),
-        # motor_parameter=dict(r_a=15e-3, r_e=15e-3, l_a=1e-3, l_e=1e-3),
-        # Take standard class and pass parameters (Load)
-        # load_parameter=dict(a=0.01, b=.1, c=0.1, j_load=.06),
-        # Pass a string (with extra parameters)
+        'DcPermExCont-v1',  # replace with 'DcPermExDisc-v1' for continuous mode
```
wallscheid, line 35: Why did you changed the control per default to continuous mode? Now together with the DC series motor example, which was already in continuous mode, we would not cover an example with discrete actions. Please arrange the two DC motor examples such that both control modes are covered by default.

examples/agents/simple_controllers.py
```diff
@@ -140,7 +182,7 @@ def __init__(self, environment, k_p=10, k_i=0.01, controller_no=0, reference_idx
         super().__init__(environment, k_p, controller_no, reference_idx)
         action_space = environment.action_space
         assert type(action_space) is Box and type(
-            environment.physical_system) is DcMotorSystem, 'No suitable action space for P Controller'
+            environment.physical_system) is DcMotorSystem, 'No suitable action space for PI Controller'
```
wallscheid, line 185: The class here is a PI controller -> hence, it should be of general form for arbitrary control problems.
In contrast to the class naming, the class content is a spezialized FOC-controller for three-phase drives which is not consistent. This specialized class may inhert from the genreal PI controller class, but this is not the case here. 

Likewise the name labeling and mapping from controller to motor classes above is a wild mix and I cannot find that starting from general controller definitions more spezialized motor-typ related controllers are derived.

examples/perm_DC_omega.py
```diff
@@ -0,0 +1,53 @@
...
+        visualization=MotorDashboard(plots=['omega', 'torque', 'i', 'u', 'u_sup'], visu_period=1),
+        ode_solver='scipy.solve_ivp', solver_kwargs=dict(),
+        reference_generator=rg.SwitchedReferenceGenerator(
+            sub_generators=[
+                rg.SinusoidalReferenceGenerator, rg.WienerProcessReferenceGenerator(), rg.StepReferenceGenerator()
+            ], p=[0.1, 0.8, 0.1], super_episode_length=(1000, 10000)
+        )
+    )
+    controller = Controller.make('p_controller', env)
+    # The above controller can be replaced with 'on_off'(disc),three_point(disc),'p_controller', 'cascaded pi' from simple_controllers class
```
wallscheid, line 42: For me it does not make much sense to use a linear P-controller to directly act on the discrete switching states of the power electronic converter. A much more natural way in order to demonstrate the usage of the motor class with discrete actions would be a hysteresis two or three point controller.

examples/pi_series_omega_control.py
```diff
@@ -1,36 +1,52 @@
...
+                   visualization=MotorDashboard(plots=['i', 'omega']),
+                   # Take standard class and pass parameters (Load)
+                   motor_parameter=dict(r_a=15e-3, r_e=15e-3, l_a=1e-3, l_e=1e-3),
+                   load_parameter=dict(a=0, b=.1, c=.1, j_load=0.04),
+                   # Pass a string (with extra parameters)
+                   ode_solver='scipy.solve_ivp', solver_kwargs={},
+                   # Pass a Class with extra parameters
+                   reference_generator=rg.WienerProcessReferenceGenerator())
+
+    controller = Controller.make('pi_controller',env)
```
wallscheid, line 48: When testing the code the PI controller resulted in frequency overcurrent situations and, therefore, leading to episode terminations. Please either change the above motor parameter in such a way that even a single PI speed controller runs with only ocassional overcurrents or rearrange the example to use a cascaded PI controller for speed and current.

examples/perm_DC_omega.py
```diff
@@ -0,0 +1,54 @@
...
+    env = gem.make(
+        'DcPermExDisc-v1',
+        visualization=MotorDashboard(plots=['omega', 'torque', 'i', 'u', 'u_sup'], visu_period=1),
+        ode_solver='scipy.solve_ivp', solver_kwargs=dict(),
+        reference_generator=rg.SwitchedReferenceGenerator(
+            sub_generators=[
+                rg.SinusoidalReferenceGenerator, rg.WienerProcessReferenceGenerator(), rg.StepReferenceGenerator()
+            ], p=[0.1, 0.8, 0.1], super_episode_length=(1000, 10000)
+        )
+    )
```
wallscheid, line 40: The visualization is not well configured: the example time step size is very small due to the discrete action space, i.e., it really takes some time to simulate just a few milliseconds. However, the standard visualization window by default shows a time intervall of 3s -> for most of the time the output window is mostly blank and only a small fraction of the let figure part is used. 

Please reconfigure the visualization options such that the depicted time intervall is more suitable to the given control problem.

wallscheid (changes requested):
- Minor changes to the examples have been directly commited by myself. Please update your branch before continue working on it.
- Some comments to the simple example code have been attached on github which should lead to easy bug fixes. Please revise the relevant code parts accordingly.  
- The simple controller class was reduced to sub-classes which at least make a minimum level of technical sense. Some classes were not functional. The relevant commits have been already pushed to the branch.
- Please fully remove the simple controllers notebook. This contains graphics inserted from external sources and does not give reference information or credits to the original authors (plagiarism). In addition, the technical descriptions of the application examples shown are in large parts either simply wrong or at least insufficient. Since there is no time for a comprehensive revision, the notebook cannot be included in the nightly/master.

examples/pmsm_foc.py
```diff
@@ -0,0 +1,64 @@
...
+    # # Change of the default motor parameters.
+    # motor_parameter = dict(
+    #     r_s=15e-3, l_d=0.37e-3, l_q=1.2e-3, psi_p=65.6e-3, p=3, j_rotor=0.06
+    # )
+    # limit_values = dict(
+    #     i=160 * 1.41,
+    #     omega=12000 * np.pi / 30,
+    #     u=450
+    # )
+    # nominal_values = {key: 0.7 * limit for key, limit in limit_values.items()}
```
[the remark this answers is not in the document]
RohithCharanD: Recommended changes have been made.

examples/perm_DC_omega.py
wallscheid: @praneeth-b : Can you assist here? 
How can Rohith change the default time interval depicted when a visualization window is opened? Currently it seems to be 3 s by default, which would be too long for the given example.

praneeth-b: The default time interval for the visualization window in the latest nightly is 1s and not 3s.
The value of 1s is because of the 'x_width' variable which is hard-coded to 1s and not parametrizable by the user at the moment. 
So the visualization is basically a sliding window of the latest 1s of the simulation and the number of points plotted is  x_width / tau = 1s / tau.
I tested  Rohit's code on his branch by setting a tau = 1e-4 and I do not see any problem in the visualization (image attached). @RohithCharanD : could you please elaborate what exactly was the problem with the visualization.?
Screenshot from 2020-10-20 21-35-38

wallscheid: mhh, I thought that the time length depicted in a window is parametrizable. Not ideal with that fixed one second length, e.g. if you want to debug and only require a couple of milliseconds to be displayed. 

Also tau will be smaller for this and similar examplars since the action space is discrete and looking from the application point of view tau = 1e-4 would be too large then. Hence, it will take some time until that 1 s window is filled with some 'signal live'.

wallscheid: @RohithCharanD : Based on the feedback of @praneeth-b  I believe we cannot resolve this issue directly within this example. 

Did you implent the other change requests so far?

RohithCharanD:
Yeah @wallscheid . The other conflicts has been resolved which were with the pmsm_foc example.

wallscheid (approved):

Issue #79, opened 2020-08-10
Calculatoric errors in induction motor initialization


max-schenke:
Nominal / Limit values of the induction motor do not get updated properly.
Initial psi_d flux can get negative.

atra94:
Closed with Pull-Request#80

PR #80, opened 2020-08-10
fixed error in flux initializer and limits updater

max-schenke:
Errors  mentioned in #79 should be fixed now.

### DIFF
--- a/gym_electric_motor/physical_systems/electric_motors.py
+++ b/gym_electric_motor/physical_systems/electric_motors.py
@@ -1716,7 +1716,7 @@ class InductionMotor(ThreePhaseMotor):
                         l_mr * u_rq_max
             psi_d_max /= - mp['p'] * omega * l_mr
             # clipping flux and setting nominal limit
-            psi_d_max = 0.9 * np.clip(psi_d_max, a_min=None, a_max=mp['l_m'] * i_d)
+            psi_d_max = 0.9 * np.clip(psi_d_max, a_min=0, a_max=np.abs(mp['l_m'] * i_d))
         # returning flux in alpha, beta system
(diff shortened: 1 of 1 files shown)


gym_electric_motor/physical_systems/electric_motors.py
```diff
@@ -1918,9 +1918,6 @@ def _update_limits(self, limit_values={}, nominal_values={}):
                                self._limits[u] / self._motor_parameter['r_s']
             nominal_agenda[i] = self._nominal_values.get('i', None) or \
                                 self._nominal_values[u] / self._motor_parameter['r_s']
-        # overwrite default limits and nominal values with func args
-        limits_agenda.update(limit_values)
-        nominal_agenda.update(nominal_values)
```
max-schenke, line 1923: I think these lines simply revert the changes made to limits_agenda, so the limits stay the same with these lines.

gym_electric_motor/physical_systems/electric_motors.py
```diff
@@ -1716,7 +1716,7 @@ def _flux_limit(self, omega=0, eps_mag=0, u_q_max=0.0, u_rq_max=0.0):
                         l_mr * u_rq_max
             psi_d_max /= - mp['p'] * omega * l_mr
             # clipping flux and setting nominal limit
-            psi_d_max = 0.9 * np.clip(psi_d_max, a_min=None, a_max=mp['l_m'] * i_d)
+            psi_d_max = 0.9 * np.clip(psi_d_max, a_min=0, a_max=np.abs(mp['l_m'] * i_d))
```
max-schenke, line 1719: I_d can be negative and the previously calculated psi_d_max can as well. This should now be caught.

PR #88, opened 2020-08-25
Example revision

max-schenke:


examples/ddpg_pmsm_dq_current_control.py
```diff
@@ -20,83 +20,138 @@
...
+        # extend the output state by zeros after reset
+        # no action can be appended yet, but the dimension must fit
         state = np.concatenate((state, np.zeros(self.env.action_space.shape)))
+
         return state, ref
 
 
 if __name__ == '__main__':
+
     tf.compat.v1.disable_eager_execution()
```
max-schenke, line 70: @atra94 
Why is it necessary to disable eager execution?
Is it only for debugging?

examples/ddpg_pmsm_dq_current_control.py
```diff
@@ -20,83 +20,138 @@
...
 
     window_length = 1
```
max-schenke, line 72: @atra94 
How can we easily explain the "window_length" that kerasRL2 uses?

examples/ddpg_series_omega_control.py
```diff
@@ -11,54 +11,93 @@
...
 from gym_electric_motor.reference_generators import WienerProcessReferenceGenerator
 from gym_electric_motor.visualization import MotorDashboard
 
+# This example shows how we can use GEM to train a reinforcement learning agent to control the motor speed
+# We use a deep-deterministic-policy-gradient (DDPG) agent to
+# determine which action must be taken on a continuous-control-set
+
 if __name__ == '__main__':
 
     tf.compat.v1.disable_eager_execution()
```
max-schenke, line 27: see above

examples/ddpg_series_omega_control.py
```diff
@@ -11,54 +11,93 @@
...
-    #  testing
+    #  CAUTION: Do not use layers that behave differently in training and testing
     #  (e.g. dropout, batch-normalization, etc..)
     #  Reason is a bug in TF2 where not the learning_phase_tensor is extractable
     #  in order to put as an input to keras models
     #  https://stackoverflow.com/questions/58987264/how-to-get-learning-phase-in-tensorflow-2-eager
     #  https://stackoverflow.com/questions/58279628/what-is-the-difference-between-tf-keras-and-tf-python-keras?noredirect=1&lq=1
     #  https://github.com/tensorflow/tensorflow/issues/34508
+
     window_length = 1
```
max-schenke, line 83: see above

examples/dqn_series_current_control.py
```diff
@@ -9,52 +9,122 @@
...
+    # For data procession we sometimes want to flatten the env output,
+    # which means that the env will only output one array that contains states and references consecutively
     env = FlattenObservation(env)
+    obs = env.reset()
 
+    # Read the number of possible actions for the given env
+    # this allows us to define a proper learning agent for this task
     nb_actions = env.action_space.n
+
     window_length = 1
```
max-schenke, line 71: see above

examples/pi_series_omega_control.py
```diff
@@ -4,42 +4,70 @@
...
-        # Pass a Class with extra parameters
+        # Define a reference generator
+        # In this example, we permit random switching between sinusoidal, wiener and step reference with
+        # probabilities 0.1, 0.8 and 0.1 respectively.
+        # super_episode_length defines minimum and maximum time steps between switching references
         reference_generator=rg.SwitchedReferenceGenerator(
             sub_generators=[
                 rg.SinusoidalReferenceGenerator, rg.WienerProcessReferenceGenerator(), rg.StepReferenceGenerator()
             ], p=[0.1, 0.8, 0.1], super_episode_length=(1000, 10000)
-        )
```
max-schenke, line 31: @atra94 
I think without knowing GEM, one is unable to see from the code that this really is supposed to be a speed control scenario. Can we make that more obvious?

max-schenke:
Some observations I made when using the dashboard:
- All the text within the plot appeared to be cursive.
- The plot for `'mean_reward'` seems to plot the first scatter point at zero reward (even before an episode has been run).

examples/ddpg_pmsm_dq_current_control.py
```diff
@@ -20,83 +20,138 @@
 from gym.core import Wrapper
 from gym.spaces import Box, Tuple
 
+# This example shows how we can use GEM to train a reinforcement learning agent to control the current within
+# a permanent magnet synchronous motor three-phase drive
```
wallscheid, line 24: I believe it would make more sense to be more precise here specialy to give all the necessary information to new users: what states are controlled in the example? 

Moreover we might just give the (trivial) information that the entire state-space is continuous.

examples/ddpg_pmsm_dq_current_control.py
```diff
@@ -20,83 +20,138 @@
...
-    observation. Therefore, for the agents it seems as if the applied action effects the state one step delayed.
-    (with a dead time of one time-step)
+    The real-world system changes its state, while the agent simultaneously calculates the next action based on a
+    previously measured observation.
+    Therefore, for the agents it seems as if the applied action affects the environment with one step delay
+    (with a dead time of one time step).
 
-    For complete observability of the system at each time-step we append the last played action of the agent to the
-    observation, because this action will be the one that is active in the next step.
+    For complete observability of the system at each time step we append the last selected action to the
```
wallscheid, line 37: The state space is completly observable within this PMSM example even without the action information. A delay is just a delay and nothing which is shadowing parts of the state space. Hence, appending the last action(s) is not related to a POMDP-style framework but just simple feature eng.

examples/ddpg_pmsm_dq_current_control.py
```diff
@@ -20,83 +20,138 @@
...
+
     tf.compat.v1.disable_eager_execution()
 
     window_length = 1
 
-    # Changing i_q reference and constant 0 i_d reference.
-    q_generator = WienerProcessReferenceGenerator(reference_state='i_sq')
-    d_generator = ConstReferenceGenerator('i_sd', 0)
+    # Define reference generators for both currents of the flux oriented dq frame
+    d_generator = ConstReferenceGenerator('i_sd', 0)  # d current reference is chosen to be constantly at zero
```
wallscheid, line 75: We should clearify:
"# d current reference is chosen to be constantly at zero just for demonstration and simplification purpose. In general, a dynamically changing d current reference is possbile and very normal for PMSM drives"

examples/ddpg_pmsm_dq_current_control.py
```diff
@@ -20,83 +20,138 @@
...
+        # Define which state variables are to be monitored concerning limit violations
+        # Here, only overcurrent will lead to termination
         observed_states=['i_sq', 'i_sd'],
+
+        # Consider converter dead time within the simulation
+        # This means that a given action will show effect only with one step delay
+        # This is realistic behavior of drive applications
         dead_time=True,
+
+        # Set the supply voltage
```
wallscheid, line 128: Set the DC-link supply voltage

examples/ddpg_pmsm_dq_current_control.py
```diff
@@ -144,24 +209,30 @@ def reset(self, **kwargs):
...
         nb_steps_warmup_actor=2048,
         nb_steps_warmup_critic=1024,
         target_model_update=1000,
         gamma=0.9,
         batch_size=128,
         memory_interval=2
     )
+
+    # Compile the models within the agent (making them ready for training)
+    # Note that the DDPG agent uses two models, hence we define two optimizers here
```
wallscheid, line 232: I would not call them "models" -> better "... two function approximators, hence..."

examples/ddpg_series_omega_control.py
```diff
@@ -11,54 +11,93 @@
...
 from gym.wrappers import FlattenObservation
-import sys, os
+import sys
+import os
 sys.path.append(os.path.abspath(os.path.join('..')))
 import gym_electric_motor as gem
 from gym_electric_motor.reference_generators import WienerProcessReferenceGenerator
 from gym_electric_motor.visualization import MotorDashboard
 
+# This example shows how we can use GEM to train a reinforcement learning agent to control the motor speed
```
wallscheid, line 21: Additional information in this abstract of the example regarding the motor type and that we are operating in a cont. state space would be nice

examples/ddpg_series_omega_control.py
```diff
@@ -11,54 +11,93 @@
...
+
+        # Define and parameterize the reference generator for the speed reference
+        reference_generator=WienerProcessReferenceGenerator(reference_state='omega', sigma_range=(5e-3, 1e-2)),
+
+        # Defines which variables to plot via the builtin dashboard monitor
+        visualization=MotorDashboard(['omega', 'torque', 'i', 'u', 'u_sup', 'reward']), visu_period=1,
     )
-    # Keras-rl DDPG-agent accepts flat observations only
+
+    # For data procession we want to flatten the env output,
```
wallscheid, line 65: data processing

examples/ddpg_series_omega_control.py
```diff
@@ -68,38 +107,44 @@
...
         target_model_update=1000,
         gamma=0.95,
         batch_size=128,
         memory_interval=1
     )
-    agent.compile((Adam(lr=1e-6), Adam(lr=1e-4)), metrics=['mae'])
 
-    # Start training for 7.5M simulation steps (1.5M training steps with actions repeated 5 times)
+    # Compile the models within the agent (making them ready for training)
+    # Note that the DDPG agent uses two models, hence we define two optimizers here
```
wallscheid, line 144: regarding "models" see above

examples/dqn_series_current_control.py
```diff
@@ -9,52 +9,122 @@
...
-import sys, os
+import sys
+import os
 sys.path.append(os.path.abspath(os.path.join('..')))
 import gym_electric_motor as gem
 from gym_electric_motor.reward_functions import WeightedSumOfErrors
 from gym_electric_motor.visualization import MotorDashboard
 from gym_electric_motor.reference_generators import WienerProcessReferenceGenerator
 
+# This example shows how we can use GEM to train a reinforcement learning agent to control the motor current
```
wallscheid, line 20: see additional information in script abstract as mentioned above

examples/dqn_series_current_control.py
```diff
@@ -9,52 +9,122 @@
...
         motor_parameter=dict(r_a=15e-3, r_e=15e-3, l_a=1e-3, l_e=1e-3),
+
+        # Set the parameters of the mechanical polynomial load (the default load class)
         load_parameter=dict(a=0, b=.1, c=.1, j_load=0.04),
-        # Pass a string (with extra parameters)
-        ode_solver='euler', solver_kwargs={},
-        # Pass a Class with extra parameters
-        reference_generator=WienerProcessReferenceGenerator(reference_state='i', sigma_range=(3e-3, 3e-2))
+
+        # Defines the utilized power converter, which determines the action space
```
wallscheid, line 37: Maybe also name it "1-phase buck converter"

examples/dqn_series_current_control.py
```diff
@@ -9,52 +9,122 @@
...
+        reference_generator=WienerProcessReferenceGenerator(reference_state='i', sigma_range=(3e-3, 3e-2)),
+
+        # Defines which variables to plot via the builtin dashboard monitor
+        visualization=MotorDashboard(plots=['i', 'omega']),
     )
+
+    # Now, the environment will output states and references separately
+    state, ref = env.reset()
+
+    # For data procession we sometimes want to flatten the env output,
```
wallscheid, line 62: data processing

wallscheid (commented):
In total, a very useful clean up and pretty up of the already available examples. Thanks @max-schenke !
Please find some smaller comments and suggestions attached to the PR.

wallscheid (approved):
Looks good from my side.

examples/ddpg_pmsm_dq_current_control.py
atra94: I think @wkirgsn  introduced this once into another example and it has been used since then. Maybe he can answer this question. Did you try to run the script without this line?

examples/ddpg_pmsm_dq_current_control.py
atra94: Those are basically just the last "window_length" observations that are used as input to the network. Therefore, not only o_k is used as input but also o_k-1, o_k-2,... . The actor and critic observation inputs are therefore shaped (window_length, observation_length)


Initially, we tried to use a longer window length instead of the AppendLastActionWrapper. However, the latter one worked better to compensate for the dead time in the system. 
Maybe we can leave this line out and hard code the value everywhere it is used.

examples/ddpg_pmsm_dq_current_control.py
wkirgsn: I can't recall what was the purpose, i would remove it if it is not necessary anymore.

Disabling eager execution would only speed things up in case it was really enabled.

max-schenke: Removal did not trigger any errors. I dropped this line.

examples/ddpg_pmsm_dq_current_control.py
max-schenke: I tried to include a short explanation of the window_length. It was used too often to hard code it, for my perception.
(3 times: actor, critic and experience replay buffer)

max-schenke: This is the appearance in the code now:
```
# Define how many past observations we want the control agent to process each step
# for this case, we assume to pass only the single most recent observation
window_length = 1
```

wkirgsn:
I have no concerns about this refurbishment, very neat and educational!

examples/userdefined_initialization.py
```diff
@@ -5,66 +5,86 @@
 from agents.simple_controllers import Controller
 import time
 
+# This code example presents how the initializer interface can be used to sample random initial states for the drive.
```
wkirgsn, line 8: Long comment blocks like these could be written with triple quotes to avoid clutter with those hashtags:
e.g.

""" This is a small
example comment
block"""

wkirgsn (approved):
good job

PR #90, opened 2020-08-28
added the plotting example to the JOSS paper

max-schenke:


max-schenke:
NOTE:
The plotting example is linked within the JOSS paper, but the link directs to the master branch. Hence, the link will not work until the example is also available on the master branch.

wallscheid (approved):

Issue #70, opened 2020-07-21
Write an example RL application using Stable Baseline3


XyDrKRulof:
Write an example application with the deep Q-network (DQN) implementation of Stable Baselines 3. Solve the current control problem of the permanent magnet synchronous motors (PMSM).

- https://github.com/DLR-RM/stable-baselines3
- Use mean absolute error (MAE) reward function
- Single discrete action variable: fundamental voltage vector with 7 options
- Cont. state variables

Take notes on the usability of the library and other notable findings. Prepare a small presentation of your results.

XyDrKRulof:
In the branch "issue_#70" is now a working version of the stable baselines3 DQN with our PMSM. Here a small view of the results:

![expected](https://user-images.githubusercontent.com/65008699/88781595-e1138180-d18c-11ea-8c74-1d122e2d29ad.png)

For the routine I did following adjustments:

- Used the hotfix provided for issue #76 in its corresponding branch
- Limit an episode length to 10000 timesteps (0.1s) to prevent overfitting on motor speeds (gym TimeLimit wrapper)
- Flattened the observation/reference Tuple to fit into stable baselines3 (gym FlattenObservation wrapper)
- Wrote an own wrapper that changes epsilon to cos(epsilon) and sin(epsilon)
- Decreased env.action_space from 8 to 7 since action 0 and 7 are equal in effect

For the hyperparameters I used the following:
- Training steps = 500000 (5s)
- Memory warmup = 10000 steps
- Memory size = 200000 steps
- Gamma = 0.99
- Learning_rate = 1e-4
- Optimizer = Adam
- Frequency of mini-batch sampling and updating: each step
- Mini-Batch-size: 25
- Network architecture: Fully connected [17, 64, 64, 7] where 17 is the flattened observation+reference and 7 is the number of actions
- Network activation functions: ReLU
- Target model update = Every 1000 steps
- Epsilon-decay: From 1 to 0.05 in the first 10% of the training steps (so it's 0.05 past 50k steps)

XyDrKRulof:
Small remark: If you want to execute the example yourself you shouldn't install stable baselines3 using pip as that version is outdated. Install it directly from their git repository (https://github.com/DLR-RM/stable-baselines3)

XyDrKRulof:
Added random state initalization at the beginning of each episode:

[...]

PR #91, opened 2020-09-02
Issue #70 and #72

XyDrKRulof:
Examples for Issue 70 and 72


max-schenke:
Can you estimate in how far the content of the upcoming tutorial notebooks will be similar to the content of this pull request?
If there are significant overlaps we should maybe drop this pull request in order to avoid redundancy in the examples.

XyDrKRulof:
Yea I guess the notebooks will contain all of the available code. We can scrap this pull request.

Issue #103, opened 2020-09-28
Bug in MultipleReferenceGenerator


atra94:
The MultipleReferenceGenerator class sets the datatype of the referenced_states array to integer. The desired datatype in all other generators is bool.

This lead to undesired behavior when automatically assigning the reward weights by the referenced states as the array indexing became faulty.

PR #104, opened 2020-09-28
Bugfix of issue#103

atra94:
Fixed the bug #103 and added a test for all reference generators to assert the correct datatype


wkirgsn (approved):
all fine

Issue #89, opened 2020-08-27
Final Dashboard Refactoring


atra94:
Last open points for the dashboard:

- Cleanup of the code
- Easier parameterization (e.g. Plot widths)
- Minor changes in class hierarchy

max-schenke:
(copied from #88)

Some observations I made when using the dashboard:

- All the text within the plot appeared to be cursive.
- The plot for 'mean_reward' seems to plot the first scatter point at zero reward (even before an episode has been run).
- A red vertical line (that marks the start of a new episode) is only plotted if the past episode has been terminated due to limit violation. Maybe it would be nice to also have a marker when the episode time limit forces the termination (maybe in a different color then).

wallscheid:
Small suggestion: 
According to @praneeth-b the time length of each sliding plotter window is hard coded to 1 second. I believe we should make this a parameter which can be modified by the user. Especially for control applications with small step sizes (e.g. single digit millisecond range for discrete actions) it will really take some time to fill up such a time length in the graphical output and also the user is not able to see the discrete actions taken at each step. In such a case it would make sense for the user to choose a much smaller time length window (maybe only 1 millisecond) to evaluate the controler behavior.

atra94:
@wallscheid of couse, this is already considered. 

The functional part of the final refactoring is almost done. "Only" extensive documentation is missing. However, to make everything really nice and clean I would need one or two additional weeks.

If the other issue regarding the tutorials is done within the next few days, I would suggest to transfer this update to the next version which could then be released much quicker than the upcoming one. If the tutorials also need some time, this issue might be solved also for the next version.

atra94:
Closed with PR

PR #114, opened 2020-10-27
Update pmsm_mpc_dq_current_control.ipynb

fbook98:


wallscheid (changes requested):
Thanks @fbook98 for the revision of the notebook. We are comming closer to a final version.
Please consider the following remarks:

- The pull request is behind the currently nightly version, so you may update you branch based on the currently GEM nightly and double check that everything works. Please also note that there is currently a major visualization update in preparation (pull request by @atra94) - so maybe you check your functionallity against the #89 branch of GEM.
- Abbreviations like 'PMSM' should be always in capital letters.
- A little text-based introduction at the beginning of the notebook would be nice, e.g. to explain what is the control task (motor & inverter type, control variables, etc.) and to shed some light on the used control solution (i,e., a brief and beginner-friendly introduction to MPC). Such an introduction can also benefit from inserting informative graphics like block diagrams.
- The following cell is not correctly described since there not only the motor is parametrized but also the reference generator for demonstration purpose. Please give more and specific information in and before that cell what is happening:
grafik
- The last cell adressing the execution of the controller is not very intuitive and informative There, at least an exemplary control scenario should be visualized: 
grafik

PR #115, opened 2020-10-28
Removal of pytest-lambda

atra94:
Everything is as discussed in issue#113

### DIFF
--- a/tests/test_physical_systems/test_mechanical_loads.py
+++ b/tests/test_physical_systems/test_mechanical_loads.py
@@ -24,7 +24,7 @@ test_bias = 10
 test_freq = 2
 
 
-def speed_profile(t, amp, freq, bias):
+def speed_profile_(t, amp, freq, bias):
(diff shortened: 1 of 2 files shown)


tests/test_physical_systems/test_mechanical_loads.py
```diff
@@ -249,21 +249,21 @@ class TestExtSpeedLoad(TestMechanicalLoad):
 
     key = 'ExtSpeedLoad'
     class_to_test = ExternalSpeedLoad
-    kwargs = dict(speed_profile=test_speed_profile,
+    kwargs = dict(speed_profile=speed_profile,
```
wkirgsn, line 252: this is no good style and poses a potential side effect.
Functions can also be keys in a dictionary, since they're also objects.

Please use a different key id or a different function name here.

wkirgsn (approved):
Besides changing the function name everything's set

tests/test_physical_systems/test_mechanical_loads.py
atra94: Done.

PR #120, opened 2020-11-11
Update of the MPC example

fbook98:


wallscheid (changes requested):
A general introduction statement is still missing, something like: "In this tutorial we apply a continuous-control set (CCS) model predictive control (MPC) approach to realize the current control of a permanent magnet synchronous motor (PMSM) in rotor field-oriented coordinates. We achieve this by utilizing the GEKKO toolbox [link]." 

In addition to the block diagram at the beginning of the notebook, the general explanation of the MPC working principle can be nicely supported by the classical lookahead picture as shown here:
https://en.wikipedia.org/wiki/Model_predictive_control#/media/File:MPC_scheme_basic.svg
(when you insert third-party figures, please provide accurate citation references)

Also figures showing the control result should only show up at the end of the notebook and not during the general introduction.

Instead of install the MPC toolbox at the beginning of the notebook, cf. 
grafik,
add it to the 'extras' requirement of the GEM toolbox, cf. https://github.com/upb-lea/gym-electric-motor/blob/master/setup.py

Please correct spelling and grammar mistakes at the headline of this cell:
grafik

The final figure of the notebook is too long in time -> since the controller is operating at 10 kHz, one can not see any details when showing a time period of 1s. Depicting a shorter time range with rich dynamic reference changes allows the reader to obtain more insights into the controller's performance.

wallscheid (approved):

wallscheid:
Wonderful tutorial, big thanks @fbook98 !

PR #121, opened 2020-11-13
Issue#102

### LINKED ISSUE #102: Clean up ConstraintMonitor
- Interfaces check-up
- general clean up

atra94:
The ConstraintMonitor update is almost ready. Only the tests and a short usage guide in the Sphinx-Docu are missing. However, a short introduction is already in the GEM-cookbook.

Before writing the tests, I would be happy to know your criticism to alter the code previously.

Closes #102 

Preliminary Release Notes on this PR:

ConstraintMonitor Refactoring:
The Constraint Monitor within the environment has been refactored to enable easy specification of complex constraints for the environments state. A guide on how to use the ConstraintMonitor can be accessed here: (Insert Link to sphinx page of CM, before next release)


examples/reinforcement_learning_controllers/ddpg_series_omega_control.py
```diff
@@ -49,8 +49,8 @@
         reward_power=0.5,
 
         # Define which state variables are to be monitored concerning limit violations
-        # None means that a limit violation will never trigger an env.reset
-        observed_states=None,
+        # () means that a limit violation will never trigger an env.reset
+        constraints=(),
```
max-schenke, line 53: I like that we are avoiding the term "observed states" in this way, as it had some error potential in the context of control theory.

max-schenke (commented):
I like the general improvements that come up in this draft:
1) the constraints are revamped to a more readable form
2) the shifted weighted reward fcn has been implemented into the "standard" weighted reward fcn

(Could you please explain which changes were applied to the juypter notebooks? That way, we don't have to open them one by one.)

Do we really want to delete the ConstraintMonitor example file? Can't we translate it / reformat it to use the new constraint interface?

Can the user still define custom constraints? It is almost certainly possible by inheriting from the now existing constraint class, but is it also possible to do it in a "simple" style as with the ConstraintMonitor? There, you were able to just define a function that receives states and returns an violation boolean.

docs/parts/reward_functions/weighted_sum_of_errors.rst
```diff
@@ -1,6 +1,36 @@
...
+    from gym_electric_motor.reward_functions import WeightedSumOfErrors
+
+
+    # initialize the reward function
+    wse = WeightedSumOfErrors(
+        reward_weights=dict(i_a=1, i_e=2) # Current control problem. Tracking of i_e is rewarded better.
+        reward_power=2 # Squared Error
+        # Alternative: reward_power=dict(i_a=1, i_e=0.5) Absolute error i_a, root error on i_e
+        bias='positive' # Shift the reward range from negative to positive
+        violation_reward=250 # Self defined violation reward
```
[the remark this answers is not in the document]
atra94: Well, the reward is automatically negated such that the resulting lv-reward is -250. However, I agree that this notation might be a bit confusing and should be changed.

atra94:
The "agent-toolbox" notebooks were only changed to fit to the new Constraint API. 

Within the GEM Cookbook notebook, I added a little introductory section about the Constraint Monitoring. Therefore, I think the extra example is not required anymore.

When deriving a custom Constraint from the Base Constraint class, you only have to implement the __call__(self, state): function. The other one set_modules(ps) is optionally. Nonetheless, I intended to make it possible to only pass a function as a Constraint. I will double check if that is possible. Otherwise the changes to make it possible are only a few lines of code.

Thank you for the quick feedback

max-schenke:
> When deriving a custom Constraint from the Base Constraint class, you only have to implement the **call**(self, state): function. The other one set_modules(ps) is optionally. 

Do we have an example of this in the GEM cookbook right now? I think if we have that we are fine.

gym_electric_motor/constraints.py
```diff
@@ -0,0 +1,96 @@
...
+        if self._observed_state_names == 'all':
+            self._observed_state_names = ps.state_names
+        if self._observed_state_names is None:
+            self._observed_state_names = []
+        self._observed_states = set_state_array(
+            dict.fromkeys(self._observed_state_names, 1), ps.state_names
+        ).astype(bool)
+
+
+class SquaredConstraint(Constraint):
```
wkirgsn, line 69: Shouldn't this inherit from LimitConstraint?

gym_electric_motor/core.py
```diff
@@ -166,10 +174,18 @@ def __init__(self, physical_system, reference_generator, reward_function, visual
         self._reward_function = instantiate(RewardFunction, reward_function, **kwargs)
         visualization = visualization or ElectricMotorVisualization()
         self._visualization = instantiate(ElectricMotorVisualization, visualization, **kwargs)
+        if isinstance(constraints, ConstraintMonitor):
+            cm = constraints
+        else:
+            limit_constraints = list(filter(lambda constraint: type(constraint) is str, constraints))
```
wkirgsn, line 180:
```suggestion
            limit_constraints = [c for c in constraints if isinstance(c, str)]
```

gym_electric_motor/core.py
```diff
@@ -755,4 +685,83 @@ def on_step_end(self):
...
+        assert merge_violations in ['max', 'product'] or callable(merge_violations)
+        self._merge_violations = lambda state: 0.0
+
+        if merge_violations == 'max':
+            self._merge_violations = max
+        elif merge_violations == 'product':
+            self._merge_violations = \
+                lambda violations: 1 - reduce(lambda val, violation: val * (1 - violation), violations, 0)
+        elif callable(merge_violations):
+            self._merge_violations = merge_violations
```
wkirgsn, line 745: assigning lambda funcs to variables is not encouraged practice.
Let's have real functions with "def" defined within this __init__ function

gym_electric_motor/core.py
```diff
@@ -755,4 +685,83 @@ def on_step_end(self):
...
+        """Function to check and merge all constraints.
+
+        Args:
+            state(ndarray(float)): The current environments state.
+
+        Returns:
+            float: The total violation degree in [0,1]
+        """
+        violations = (constraint(state) for constraint in self._constraints)
+        return self._merge_violations(violations)
```
wkirgsn, line 767:
```suggestion
        return self._merge_violations(*violations)

```

wkirgsn (approved):
Again very minor remarks, refactoring successful

examples/classic_controllers/perm_dc_omega.py
```diff
@@ -47,7 +50,7 @@
     for i in range(100000):
         env.render()
         action = controller.control(state, reference)
-        (state, reference), reward, done, _ = env.step(action)
+        (state, reference), reward, done, _ = env.step(1)
```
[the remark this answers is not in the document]
atra94: Sorry, I modified it for a quick test to force a limit violation. This was not intended to be committed. I will undo it.

gym_electric_motor/constraints.py
atra94: There was no real reuse of code possible, imo. Maybe this becomes interesting if the number of available Contraints grows, but I don't expect that at the moment.

PR #123, opened 2020-12-13
Issue#111 closes #111 

### LINKED ISSUE #111: Notebook visu is not appearing during agent testing
In the [keras-rl2](https://colab.research.google.com/github/upb-lea/gym-electric-motor/blob/master/examples/reinforcement_learning_controllers/keras_rl2_dqn_disc_pmsm_example.ipynb) notebook, visualization plots are not created during the testing phase. 

They appeared in an older version of the notebook, where a new environment was instantiated after training.
However, it shouldn't be necessary to instantiate new environments to have a plot for successive testing loops.

atra94:
Fixed the problems regarding the plotting in jupyter notebooks.

- With the method MotorDashboard.reset_figures(), the dashboard can be reset in a way, that it is ready for a new plot in a new notebook cell.


gym_electric_motor/visualization/motor_dashboard.py
```diff
@@ -212,7 +250,15 @@ def _initialize(self):
             for plot, axis in zip(self._step_plots, axes_int):
                 plot.initialize(axis)
 
-        plt.pause(0.1)
+        if len(self._time_plots) > 0:
+            self._time_plot_figure, axes_step = plt.subplots(len(self._time_plots), sharex=True)
+            self._time_plot_figure.canvas.set_window_title('Time Plots')
+            axes_step = [axes_step] if len(self._time_plots) == 1 else axes_step
```
wkirgsn, line 256: Is this robust in so far that the len Attribute will always be present?

gym_electric_motor/visualization/motor_dashboard_plots/reward_plot.py
```diff
@@ -30,6 +30,10 @@ def set_env(self, env):
         self._y_lim = (min_limit - spacing, max_limit + spacing)
         self._label = 'reward'
 
+    def reset_data(self):
+        super().reset_data()
+        self._reward_data = np.zeros_like(self._x_data, dtype=float) * np.nan
```
wkirgsn, line 35: Is this producing zeros or nans?

wkirgsn (commented):
Static Code review shows decent code. Runtime Tests and Jup notebook testing pending (on my side)

gym_electric_motor/visualization/motor_dashboard.py
atra94: Yes. The self._time_plots are set as a list within the class and not as a user argument. Therefore, the len() argument is safe.

gym_electric_motor/visualization/motor_dashboard_plots/reward_plot.py
atra94: This is producing nans. Do you know a more elegant way to initialize a nan-array? Nans are used because they generate a non visible line when passed to matplotlib. When the env simulates, the nans are overwritten with the reward signal.

wkirgsn: Checkout np.full

https://numpy.org/doc/stable/reference/generated/numpy.full.html

atra94: Fixed 😃

wkirgsn (approved):
after merge, the Tutorial notebooks should be checked on Colab

PR #125, opened 2021-01-21
Issue 124 dc perm ex

atra94:
An intermediate PR only for a review on the changes on the PermExDC Motor. 

The other environments will be designed accordingly.


max-schenke:
Looks good upon visual code inspection, maybe consider to deliver some executable example.

In the scope of the planned update we should discuss whether we want to keep the env names containing "cont" and "disc".
Of course we need to make this distinction, but I would like to change the naming from "disc" to "finite" (or an abbreviation).
In drive control the distinction is usually made between Continuous-Control-Set (CCS) and Finite-Control-Set (FCS) methods, so we could align with this naming convention. This update could be a good opportunity to do that.

Later we might want to add "hybrid" (?) control scenarios, where the controller needs to choose both, a switching signal from a finite set and a switching time from a continuous set (obviously not in this update).

atra94:
When all environments are updated, I will also work on the examples and especially the cookbook to introduce all changes.

If all agree that the wording "Finite" fits better than "Discrete", then we should fix it now. 
Maybe we could directly change to the abbreviations FCS / CCS such that the environment ids would be structured as follows:

{FCS/CCS}-{TC/SC/CC}-{MotorType}-v0

In the long run, also Hybrid Controlled Environments could be introduced.

PR #130, opened 2021-02-16
Bugfix in Readme.md

atra94:
Changed naming of mpc-example in README.md and extended minimal example by reset after done


wkirgsn (approved):
all ok

PR #132, opened 2021-03-18
WIP: Issue 124

atra94:
The Environment Refactoring is almost done. Only minor points are still open. However, I would like to receive an early review to introduce your feedback and finally close this issue.

Open Points:

- Little documentation update of the induction motors
- Writing of basic environment unit tests

Change Notes

- LaplaceReferenceGenerator: Equivalent to the WienerProcess References, a Laplace Reference Generator was developed that uses the Laplace Distribution as backend random number generating process
- Ornstein Uhlenbeck Load: An externally forced, dynamic load process was delevoped that is based on the OU-Process
- Standardized environments were introduced that can be used out-of-the-box without any neccessary parameterization.


wkirgsn:
This seems to be a hefty update, it will take some time to review.
I hope to get to it by next week. In the mean time, maybe check why travis is not building?

atra94:
> 
> 
> This seems to be a hefty update, it will take some time to review.
> I hope to get to it by next week. In the mean time, maybe check why travis is not building?

Every environment has got basically the same changes. This should reduce the review effort a lot. Furthermore, I merged a pre-release version of the classic controller update to test the environments. This code should be reviewed, when the classic controllers are finished.

The failing tests are due to changes in the motor parameters in the electric motor tests. However the tests are absolutely not maintainable. I try to avoid rewriting these completely but I dont know if that is possible.

gym_electric_motor/utils.py
```diff
@@ -56,6 +56,18 @@ def set_state_array(input_values, state_names):
     return state_array
 
 
+def initialize(base_class, arg, default_class, default_args):
+    if arg is None:
+        return default_class(**default_args)
+    elif isinstance(arg, base_class):
+        return arg
+    elif type(arg) is str:
```
wkirgsn, line 64: rather use isinstance()

wkirgsn (changes requested):
Thank you for your updates.
Some implementations seem a bit prototypic, there I left appropriate comments.
If something is annotated as a general remark, please adapt all corresponding places in the code.

wkirgsn:
I am not sure if I got you right @atra94.
If major changes break unit tests, then we need to have the tests updated before we can merge everything to master/nightly

atra94:
> 
> 
> I am not sure if I got you right @atra94.
> If major changes break unit tests, then we need to have the tests updated before we can merge everything to master/nightly

The problem is, that only an "update" will not fix the test. Probably, completely new tests from scratch might be faster and more readable in the future.

But that will still take some time.

gym_electric_motor/envs/gym_im/doubly_fed_induction_motor_envs/finite_sc_dfim_env.py
```diff
@@ -0,0 +1,165 @@
...
+from gym_electric_motor import physical_systems as ps
+from gym_electric_motor.reward_functions import WeightedSumOfErrors
+from gym_electric_motor.utils import initialize
+from gym_electric_motor.constraints import SquaredConstraint
+
+
+class FiniteSpeedControlDoublyFedInductionMotorEnv(ElectricMotorEnvironment):
+    """
+    Description:
+        Environment to simulate a dq-domain continuous control set current controlled doubly fed induction motor.
```
wkirgsn, line 15: This class and file talks about finite control and the description here speaks of continuous control.
Moreover, the key in the description talks about "AbcCont" but the first line speaks of dq-domain.
Please check for all other classes for correct documentation compliance.

wkirgsn: I highly suggest to write a program to check and even replace text in these files to be conform with the class names and file names

gym_electric_motor/physical_systems/electric_motors/dc_motor.py
```diff
@@ -0,0 +1,155 @@
...
+        # Docstring of superclass
+        super().__init__(motor_parameter, nominal_values, limit_values, motor_initializer)
+        #: Matrix that contains the constant parameters of the systems equation for faster computation
+        self._model_constants = None
+        self._update_model()
+        self._update_limits()
+
+    def _update_model(self):
+        """
+        Update the motors model parameters with the motor parameters.
```
wkirgsn, line 80: motor's

wkirgsn (approved):
except for some wording everything looks good

PR #136, opened 2021-06-11
Classic controller

fbook98:


examples/classic_controllers/classic_controllers.py
```diff
@@ -380,21 +497,34 @@ def reset(self):
...
         self.i_idx = environment.physical_system.CURRENTS_IDX[-1]
         self.control_e = type(environment.physical_system.electrical_motor) == DcExternallyExcitedMotor
+        self.state_names = environment.state_names
+
+        self.external_ref_plots = external_ref_plots
+        for ext_ref_plot in self.external_ref_plots:
+            ext_ref_plot.set_reference(ref_states)
 
         if self.control_e:
             assert len(stages) == 2, 'Controller design is not completely'
```
max-schenke, line 526: 'Controller design is incomplete.'

examples/classic_controllers/classic_controllers.py
```diff
@@ -688,41 +902,507 @@ def control(self, state, reference):
...
+        self.mp = environment.physical_system.electrical_motor.motor_parameter
+        self.limit = environment.physical_system.limits
+        self.nominal_values = environment.physical_system.nominal_state
+        self.torque_control = torque_control
+
+        self.q = environment.physical_system.electrical_motor.q
+        self.a_max = 2 / np.sqrt(3)
+        self.k_ = 0.95
+        d = 1.2
+        alpha = d / (d - np.sqrt(d ** 2 - 1))
```
max-schenke, line 978: These variables (and their assigned values) are probably not very self-explanatory. They could use some in-line commenting and / or more descriptive names.

max-schenke (commented):
Thank you for your effort! 
Upon code inspection I only found minor points to be addressed. 

Please also have a second look at the docstrings of each class / function. 

While plotting, the labels should be written in italic font ("latex math mode") if the letter denotes a variable (e.g. torque T, current i, voltage u etc.). Abbreviations and units should be written in upright font, e.g., $i_\mathrm{d,max} / \mathrm{A}$ instead of  $i_{d,max} / A$. (I remember that the $\text{}$ command does not work in the matplotlib-latex combination, but maybe this has been patched by now.)

max-schenke:
I get a warning on each line that deals with latex-based strings. I think the solution to this is to mark the string as "raw". 
E.g., `label = r"$\Psi_e$"` instead of `label = "$\Psi_e$"` (notice the "r" in front of the string).
(I am using Windows, maybe this is less of an issue on Linux?)

examples/classic_controllers/classic_controllers.py
```diff
@@ -2,16 +2,39 @@
 from gym_electric_motor.physical_systems import SynchronousMotorSystem, DcMotorSystem, DcSeriesMotor, \
     DcExternallyExcitedMotor
 from gym_electric_motor.reference_generators import MultipleReferenceGenerator, SwitchedReferenceGenerator
-from gym_electric_motor.envs import ContDcExternallyExcitedMotorEnvironment, DiscDcExternallyExcitedMotorEnvironment, ContPermanentMagnetSynchronousMotorEnvironment
+from gym_electric_motor.visualization import MotorDashboard
+from gym_electric_motor import envs
```
max-schenke, line 6: As it seems this import does not make use of the "local" GEM files within the same directory but rather necessitates that GEM is installed in the used Python environment. 
This complicates testing of new code and should therefore be avoided.

max-schenke: (If this is the case it applies to all files of this pull request.)

max-schenke:
Naming the given files "test_..." (test_classic_controllers_dc_motor.py, test_classic_controllers_synch_motor.py) seems to cause unintended behavior if pytest is installed because it will treat them as unit tests with a different execution protocol. This should be fixed by renaming the files.

examples/classic_controllers/externally_referenced_state_plot.py
```diff
@@ -2,47 +2,16 @@
 
 
 class ExternallyReferencedStatePlot(StatePlot):
-    """Plot that displays environments states together with externally generated references.
```
atra94, line 5: Maybe you can just re-add the documentation here, if it still fits.

examples/classic_controllers/classic_controllers_synch_motor_test.py
```diff
@@ -3,95 +3,50 @@
 import gym_electric_motor as gem
```
atra94, line 3: If this file should be added to the nightly/master branch, it should be tidied up a little

atra94 (changes requested):
Nice work 👍 . I mentioned a lot of points, but they are all minor changes and most of all only code style.

examples/classic_controllers/classic_controllers.py
```diff
@@ -632,10 +800,17 @@ def __init__(self, environment, stages, ref_states, color='g', **controller_kwar
...
+            self.ref_d_idx = np.where(ref_states == 'i_sd')[0][0]
+            self.ref_idx = np.where(ref_states != 'i_sd')[0][0]
+            self.ref_state_idx = [self.i_sq_idx, environment.state_names.index(ref_states[self.ref_idx])]
 
-        self.ref_d_idx = np.where(ref_states == 'i_sd')[0][0]
-        self.ref_idx = np.where(ref_states != 'i_sd')[0][0]
-        self.ref_state_idx = [self.i_sq_idx, environment.state_names.index(ref_states[self.ref_idx])]
         cont_pmsm_envs = (
             envs.DqContCurrentControlPermanentMagnetSynchronousMotorEnv,
             envs.DqContTorqueControlSquirrelCageInductionMotorEnv,
```
wallscheid, line 816: Why is here an induction machine part of the PMSM env set?

examples/classic_controllers/classic_controllers.py
```diff
@@ -704,41 +902,507 @@ def control(self, state, reference):
...
+            self.q_controller.reset()
+
         else:
             for abc_controller in self.abc_controller:
                 abc_controller.reset()
 
+        if self.torque_control:
+            self.torque_controller.reset()
+
+class MTPC:
```
wallscheid, line 952: I feel that the class name 'MTPC' is not fitting well to the actual functionality, since the class additionally incorporates the MTPF characteristics for flux weakening mode. A better name would ne 'OPS' (for Operationg point selection) or 'TorqueToCurrentConversion' or similar.

examples/classic_controllers/classic_controllers.py
```diff
@@ -704,41 +902,507 @@ def control(self, state, reference):
...
+                    i_q = np.sign(torque) * np.abs(self.mtpc[psi_idx_, 2])
+        if i_d < self.mtpf[psi_max_idx, 2]:
+            i_d = self.mtpf[psi_max_idx, 2]
+            i_q = np.sign(torque) * np.abs(self.mtpf[psi_max_idx, 3])
+
+        if self.plot_torque:
+            if self.k == 0:
+                self.intitialize_torque_plot()
+
+            self.k_list.append(self.k / 10000)
```
wallscheid, line 1287: too many data points -> also use a configuration variable for it

examples/classic_controllers/classic_controllers.py
```diff
@@ -704,41 +902,507 @@ def control(self, state, reference):
...
+
+        else:
+            psi_delta = np.clip(psi_delta, self.psi_low, self.psi_high)
+            self.limited = True
+
+        psi = psi_max_ + psi_delta
+        if self.plot_modulation:
+            if self.k == 0:
+                self.initialize_modulation_plot()
+            self.k_list_a.append(self.k / 10000)
```
wallscheid, line 1350: see above

examples/classic_controllers/classic_controllers.py
```diff
@@ -704,41 +902,507 @@ def control(self, state, reference):
...
+            self.a_list.append(a)
+            self.psi_delta_list.append(psi_delta)
+
+            if self.k % self.update_interval == 0:
+                    self.a_plot.scatter(self.k_list_a, self.a_list, c='tab:blue', s=3)
+                    self.psi_delta_plot.scatter(self.k_list_a, self.psi_delta_list, c='tab:blue', s=3)
+                    self.a_plot.set_xlim(max(self.k / 10000, 1) - 1, max(self.k / 10000, 1))
+                    self.psi_delta_plot.set_xlim(max(self.k / 10000, 1) - 1, max(self.k / 10000, 1))
+                    self.k_list_a = []
```
wallscheid, line 1359: see above

examples/classic_controllers/classic_controllers_synch_motor_test.py
```diff
@@ -3,95 +3,50 @@
...
-            rg.StepReferenceGenerator(reference_state='i_sd', amplitude_range=(0, 0.3), offset_range=(0, 0.2)),
-            rg.SinusoidalReferenceGenerator(reference_state='i_sd', amplitude_range=(0, 0.3), offset_range=(0, 0.2)),
-            rg.ConstReferenceGenerator(reference_state='i_sd', value=0)],
-        p=[0.25, 0.1, 0.25, 0.2, 0.2], super_episode_length=(1000, 10000)
-    )
-    '''
-    d_generator = rg.ConstReferenceGenerator(reference_state=ref_states[0], reference_value=0)
+    
+    external_ref_plots = [ExternallyReferencedStatePlot(state) for state in ['omega', 'torque', 'i_sd', 'i_sq', 'u_sd', 'u_sq']]
+    matplotlib.use('qt5agg')
```
wallscheid: can be deleted, also raised an error on my machine

examples/classic_controllers/classic_controllers_synch_motor_test.py
```diff
@@ -3,95 +3,50 @@
...
-            rg.WienerProcessReferenceGenerator(reference_state='i_sd'),
-            rg.StepReferenceGenerator(reference_state='i_sd', amplitude_range=(0, 0.3), offset_range=(0, 0.2)),
-            rg.SinusoidalReferenceGenerator(reference_state='i_sd', amplitude_range=(0, 0.3), offset_range=(0, 0.2)),
-            rg.ConstReferenceGenerator(reference_state='i_sd', value=0)],
-        p=[0.25, 0.1, 0.25, 0.2, 0.2], super_episode_length=(1000, 10000)
-    )
-    '''
-    d_generator = rg.ConstReferenceGenerator(reference_state=ref_states[0], reference_value=0)
+    
+    external_ref_plots = [ExternallyReferencedStatePlot(state) for state in ['omega', 'torque', 'i_sd', 'i_sq', 'u_sd', 'u_sq']]
```
wallscheid, line 18: As mentioned above, we should display distinctly different use cases of the synchronous motor control part and the reference signals shown in each use case should fit the given scenario. For example, the users get a speed reference signal displayed although torque control mode is activated which is likely to confuse them.

examples/classic_controllers/classic_controllers_synch_motor_test.py
```diff
@@ -3,95 +3,50 @@
...
+        #load=ExternalSpeedLoad(lambda t: ((np.sin(2 * np.pi * 5 * t) + 1) * 2e3 + 5e3) * np.pi / 30),
+        load=PolynomialStaticLoad(),
         reference_generator=reference_generator,
     )
 
-    controller = Controller.make(env)
+    controller = Controller.make(env, external_ref_plots=external_ref_plots, torque_control='online')
     state, reference = env.reset()
-
     steps = 10001
```
wallscheid, line 39: same plotting issues as with the DC motors: t-time axis is not adjusted and the plotting / calculation process is not correctly terminated / finished after the step number is reached.

wallscheid (changes requested):
Very nice work – good progress!

Besides my comments within the code, where some of them maybe related more to the plotting routines than the examples given here (maybe @atra94  can have a view on them too), I have the following comments:

-   the files for the classic controller examples should be named ‘example’ and not ‘test’.
-	All old files related to the ‘simple controller class’ (incl. examples) should be deleted since they are largely redundant (and of much poorer quality)
-	I would like to see at least one additional example for the DC motors and one for the PMSM where a manual overwrite / design of the controller values is demonstrated – this functionality is currently not really highlighted (only automated tuning is used so far).

examples/classic_controllers/classic_controllers_dc_motor_test.py
```diff
@@ -1,77 +1,56 @@
...
-    steps = 10001 if type(env.action_space) == Box else 100001
+    controller = Controller.make(env, external_ref_plots=external_ref_plots)
+    steps = 10001
     state, reference = env.reset()
-    cum_rew = 0
 
     for i in range(steps):
-
         action = controller.control(state, reference)
-        env.render(ref=controller.plot())
```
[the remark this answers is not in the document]
atra94: The problem here is, that the controller needs only a very little average voltage to hit the current reference line. 
Probably, only one or two control cycles a "voltage high" action is required to drive the current to the reference.

However, the scatter plot points are quite tiny and almost invisible. The usage of line plots as alternative would result in very cluttered (for all quantities except the speed) and is also a lot slower to plot online. 

I did not really find a good solution. Thick points are also visible on plots like this one. On the contrary, they also blur out details in the plots. 
In the next days, I will have a look at this topic again.

examples/classic_controllers/classic_controllers_dc_motor_test.py
```diff
@@ -1,77 +1,56 @@
...
-
-        reference_generator=reference_generator
-    )
-    controller = Controller.make(env)
+    matplotlib.use('TkAgg')
+    env = gem.make(motor, visualization=MotorDashboard(additional_plots=external_ref_plots))
 
-    steps = 10001 if type(env.action_space) == Box else 100001
+    controller = Controller.make(env, external_ref_plots=external_ref_plots)
+    steps = 10001
```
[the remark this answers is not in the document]
atra94: The default time-Axis width in the MotorDashboard are 10000 steps. If you simulate less than 10k steps, the window will stay partly blank. However, the window should slide, if you simulate for a longer time. 

If desired, the time-Axis width can be adjusted by the user.

examples/classic_controllers/classic_controllers_dc_motor_test.py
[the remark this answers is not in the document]
atra94: grafik
How about this? As mentioned, the details are blurred out, but now you can see single points. Furthermore, I reduced the thickness of the limit lines a little

examples/classic_controllers/classic_controllers.py
```diff
@@ -389,21 +497,34 @@ def reset(self):
...
-        psi_e = self.psi_e or self.l_e * state[self.i_idx] * self.limit[self.i_idx]
-        return (state[self.omega_idx] * self.limit[self.omega_idx] * psi_e) / self.limit[self.u_idx]
+        psi_e = self.psi_e or self.l_e * state[self.i_idx] * self.nominal_values[self.i_idx]
+        return (state[self.omega_idx] * self.nominal_values[self.omega_idx] * psi_e) / self.nominal_values[self.u_idx]
 
 
 class DiscreteActionController(Controller):
-    def __init__(self, environment, stages, ref_states, **controller_kwargs):
+    """
+        This class is used for current control of all discrete DC motor systems. By default, a three-point controller is
```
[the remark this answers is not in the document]
atra94: Done. I just piggybacked it to this Pull Request.  😅

examples/classic_controllers/classic_controllers.py
```diff
@@ -231,10 +319,12 @@ def automated_gain(environment, stages, controller_type, **controller_kwargs):
                                     'd_gain']
 
                         elif i == 1:
+                            p_gain_ = mp['l'] / (environment.physical_system.tau * a) / u_a_lim * i_a_lim
+                            i_gain_ = p_gain_ / (environment.physical_system.tau * a ** 2)
+                            t_n = p_gain_ / i_gain_
```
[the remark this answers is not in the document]
fbook98: I have changed the calculation of t_n (t_n = tau * a²).

examples/classic_controllers/classic_controllers.py
```diff
@@ -297,9 +387,11 @@ def automated_gain(environment, stages, controller_type, **controller_kwargs):
                         d_gain_q = p_gain_q * environment.physical_system.tau
                         stage_q[0]['d_gain'] = d_gain_q if 'd_gain' not in stage_q[0].keys() else stage_q[0]['d_gain']
 
-                    if ('omega' in ref_states or 'torque' in ref_states) and _controllers[stage_q[1]['controller_type']][1] == ContinuousController:
-                        p_gain = environment.physical_system.mechanical_load.j_total / (1.5 * a ** 2 * mp['p'] * np.abs(mp['l_d'] - mp['l_q'])) / i_sq_lim * omega_lim
-                        i_gain = p_gain / (1.5 * environment.physical_system.tau * a)
+                    if 'omega' in ref_states and _controllers[stage_q[1]['controller_type']][1] == ContinuousController:
+                        t_n = p_gain_d / i_gain_d
+                        p_gain = environment.physical_system.mechanical_load.j_total / (4 * t_n) / torque_lim * omega_lim
+                        i_gain = p_gain / (4 * t_n)
```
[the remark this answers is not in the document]
fbook98: For p_gain the factor must be a**2 and for i_gain a.

examples/classic_controllers/classic_controllers_dc_motor_test.py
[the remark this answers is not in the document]
wallscheid: in my case, it did not slide for more steps -> time length was always fixed to 0.1 seconds in all dc and PMSM use cases I have tested. 

Isn't it possible to formulate a maximum step number which is depicted in a figure, and if the simulated step number is lower that the t axis is reduced automatically?

examples/classic_controllers/classic_controllers_dc_motor_test.py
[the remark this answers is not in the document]
wallscheid: a little better

examples/classic_controllers/classic_controllers_dc_motor_test.py
```diff
@@ -1,77 +1,56 @@
...
-            sub_generators=[
-                rg.SinusoidalReferenceGenerator(reference_state=ref[0], amplitude_range=(0, 0.3), offset_range=(0, 0.2)),
-                rg.WienerProcessReferenceGenerator(reference_state=ref[0], amplitude_range=(0, 0.3),
-                                                   offset_range=(0, 0.2)),
-                rg.StepReferenceGenerator(reference_state=ref[0], amplitude_range=(0, 0.3), offset_range=(0, 0.2))
-            ], p=[0.5, 0.25, 0.25], super_episode_length=(10000, 100000)
-        )
+    """
+    motor type:     'PermExDc'  Permanently Excited DC Motor
+                    'ExtExDc'   Externally Excited MC Motor
```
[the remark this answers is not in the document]
fbook98: I have added the reference signal to the plot.

examples/classic_controllers/classic_controllers.py
fbook98: I have added in-line comments.

examples/classic_controllers/classic_controllers_dc_motor_test.py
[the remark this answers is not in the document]
atra94: During the last Update of the Dashboard, I already experimented with  "growing" plots.
To me, it looked confusing because the points were moving at each plot update.

Furthermore, a simulation of less then 10k points is only for test cases. And if you keep the figure open with the plt.show(block=True) command, you can zoom to the desired window.

The problem that your plot is not sliding originates probably from your default matplotlib backend. Which one do you typically use? And can you try to execute a script from an (anaconda) shell instead of from PyCharm?

examples/classic_controllers/classic_controllers_dc_motor_test.py
```diff
@@ -1,77 +1,56 @@
...
+                    
+    modelling:      'Cont'      Continuous Action Space
+                    'Finite'    Discrete Action Space
+    """
+
+    motor_type = 'SeriesDc'
+    control_type = 'T'
+    modelling = 'Cont'
+
+    motor = modelling + '-' + control_type + 'C-' + motor_type + '-v0'
```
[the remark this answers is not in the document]
fbook98: I renamed the variable to 'env_id' as it is named in gem_control.

wkirgsn:
I would add my review after the so-far requested changes were taken care of as there is quite a bit to adapt already

examples/classic_controllers/classic_controllers.py
```diff
@@ -704,41 +902,507 @@ def control(self, state, reference):
...
+
+        self.k += 1
+
+        return i_q, i_d
+
+    def modulation_control(self, state):
+        eps = state[self.epsilon_idx] * self.limit[self.epsilon_idx]
+        u_sd = state[self.u_sd_idx] * self.limit[self.u_sd_idx]
+        u_sq = state[self.u_sq_idx] * self.limit[self.u_sq_idx]
+        u_alpha, u_beta = self.q([u_sd, u_sq], eps)
```
atra94, line 1337: @wallscheid, I just wondered, if the transformation to alpha-beta coordinates is required at this point. 
It is only used to calculate the modulation degree "a" with the length of the voltage vector. 

As only circular voltage set points are used here and the hexagon is neglected, I think this is redundant.

examples/classic_controllers/classic_controllers.py
```diff
@@ -704,41 +902,507 @@ def control(self, state, reference):
...
+        a = 2 * np.sqrt(u_alpha ** 2 + u_beta ** 2) / self.u_dc
+
+        if a > 1.1 * self.a_max:
+            self.integrated = self.integrated_reset
+
+        a_delta = self.k_ * self.a_max - a
+        if state[self.omega_idx] > 0.0001:
```
atra94, line 1344: In my opinion, it should be `np.abs(state[self.omega_idx])`. Currently, it enters the else clause for every negative speed value. Or is this intended? Furthermore, I think you can reduce the if-else clause completely by just writing sth. like the following:
```
# Avoid division by zero at 0 speed
omega = max(np.abs(state[self.omega_idx]), 0.001)
psi_max_ = self.u_dc / (np.sqrt(3) * self.limit[self.omega_idx] * omega * self.p)
```

examples/classic_controllers/classic_controllers.py
wallscheid: Yes, I agree. The voltage vector length used in the following can be directly calculated from dq coordinates.

fbook98: I have changed the calculation of a.

examples/classic_controllers/classic_controllers.py
```diff
@@ -389,21 +497,34 @@ def reset(self):
...
+
+        self.external_ref_plots = external_ref_plots
+        for ext_ref_plot in self.external_ref_plots:
+            ext_ref_plot.set_reference(ref_states)
 
         if self.control_e:
             assert len(stages) == 2, 'Controller design is not completely'
+
             self.ref_e_idx = False if 'i_e' not in ref_states else np.where(ref_states == 'i_e')[0][0]
             self.ref_e = 0.1 if 'ref_e' not in controller_kwargs.keys() else controller_kwargs['ref_e']
```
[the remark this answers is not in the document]
fbook98: I have added a Torque to Current controller for the ExtExDc.

examples/classic_controllers/classic_controllers.py
```diff
@@ -648,97 +865,613 @@ def __init__(self, environment, stages, ref_states, color='g', **controller_kwar
...
+            self.psi_delta_plot.set_xlim(0, 1)
+            self.psi_delta_plot.legend(loc=2)
+
+            self.a_list = []
+            self.psi_delta_list = []
+            self.k_list_a = []
+
+
+    def reset(self):
+        self.integrated = 0
```
wkirgsn, line 1469:
```suggestion
        self.integrated = self.integrated_reset
```
seems more reasonable?

examples/classic_controllers/classic_controllers.py
```diff
@@ -772,6 +1505,12 @@ def __init__(self, d_gain=0.05):
 
 
 class PI_Controller(P_Controller, I_Controller):
+    """
+        The PI-Controller is a combination of the base P-Controller and the base I-Controller. The integrate function
```
wkirgsn, line 1509: sentence incomplete

examples/classic_controllers/classic_controllers.py
```diff
@@ -648,97 +865,613 @@ def __init__(self, environment, stages, ref_states, color='g', **controller_kwar
...
+            self.i_q_inter_plot = self.i_q_inter.T
+
+        elif torque_control == 'interpolate':
+            self.t_grid, self.psi_grid = np.mgrid[np.amin(t):np.amax(t):np.complex(0, self.t_count),
+                                                  self.psi_min:self.psi_max:np.complex(self.psi_count)]
+            self.i_q_inter = griddata((t, psi), i_q, (self.t_grid, self.psi_grid), method='linear')
+            self.i_d_inter = griddata((t, psi), i_d, (self.t_grid, self.psi_grid), method='linear')
+            self.i_d_inter_plot = self.i_d_inter
+            self.i_q_inter_plot = self.i_q_inter
+
```
wkirgsn, line 1208: since `torque_control` is user input, maybe add an eventual else clause here raising NotImplementedError for strings different from "analytical" and "interpolate"

wkirgsn (changes requested):
Some minor style remarks, nothing critical.

examples/classic_controllers/classic_controllers.py
```diff
@@ -339,48 +430,73 @@ def automated_gain(environment, stages, controller_type, **controller_kwargs):
...
                                                                                  **controller_kwargs)
 
     def control(self, state, reference):
         self.action[0] = self.controller.control(state[self.ref_state_idx], reference[self.ref_idx]) + self.feedforward(
             state)
-        if self.action_space.low[0] <= self.action[0] <= self.action_space.high[0]:
+        if self.action_limit_low <= self.action[0] <= self.action_limit_high:
             self.controller.integrate(state[self.ref_state_idx], reference[self.ref_idx])
+        else:
+            self.action[0] = np.clip(self.action[0], self.action_limit_low, self.action_limit_high)
```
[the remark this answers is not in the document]
fbook98: It is not integrated to prevent a wind up.

wkirgsn:
@fbook98 if you are done applying change requests, please re-request a review at all reviewers

examples/classic_controllers/classic_controllers_synch_motor_example.py
```diff
@@ -0,0 +1,69 @@
...
+                        'T'         Torque Control
+                        'C'         Current Control
+
+        action_type:    'AbcCont'   Continuous Action Space in ABC-Coordinates
+                        'Finite'    Discrete Action Space
+    """
+
+    motor_type = 'PMSM'
+    control_type = 'S'
+    action_type = 'AbcCont'
```
wallscheid, line 23: The following combinations resulted in an error:

- PMSM + S + Finite  (issue with automated controller staging)
- PMSM + C + Finite (issue with automated controller staging)
- All SynRm options (no registered env. error)

 Please resolve these issues and/or add a comment to this example file which configurations are actually working at the moment

examples/classic_controllers/classic_controllers_synch_motor_manual_example.py
```diff
@@ -0,0 +1,113 @@
...
+                        
+        control type:   'S'         Speed Control
+                        'T'         Torque Control
+                        'C'         Current Control
+
+        action_type:    'AbcCont'   Continuous Action Space in ABC-Coordinates
+                        'Finite'    Discrete Action Space
+    """
+
+    motor_type = 'PMSM'
```
wallscheid, line 21: All SynRm variants resulted in an 'no registered env' type of error.

wallscheid (changes requested):
Some minor issues with some of the three-phase machine examples remain - please have a look. Except that, I believe that the pull request / classic controller suite is in a very good shape - good work!

examples/classic_controllers/classic_controllers_synch_motor_example.py
atra94: The key for the synchronous reluctance motor is SynRM and not SynRm. With this key the environments are executable.#
The staging error appears for the finite control environments of this motor type, too.

Furthermore, the control performance looks not satisfying to me.

There are multiple current limit violations in the current control env. I think I should have a look at the motor parameters here. The current limit is too close to the nominal value of the currents.

Additionaly, there is a steady state error in the torque control application. 

Finally, the reference of the speed control environment is still too dynamic. However, I think the controller could also act a little more aggressive.

examples/classic_controllers/classic_controllers_synch_motor_manual_example.py
wkirgsn: @fbook98 Let's have a cast to lower on all string user inputs with ```s.lower()``` and compare with the corresponding expected lowercase strings

atra94: @wkirgsn, I dont think that we can do anything here (at least regarding the environment ids). This would require changes in the registration and creation of environments within code of OpenAI-GYM.

examples/classic_controllers/classic_controllers_synch_motor_example.py
fbook98: I have found the bug of the synchronous motors. I will have a look on the torque control of the SynRM.

fbook98: It was just an error in the manual definition of the environment. A psi_p was also passed for the SynRM, which caused the torque control error. The SynRM also had an error with the current limitation of the torque control. This has also been fixed.

wallscheid (approved):
Everything fine from my point of view - thanks for the changes.

examples/classic_controllers/custom_classic_controllers_synch_motor_example.py
```diff
@@ -0,0 +1,114 @@
...
+        control type:   'S'         Speed Control
+                        'T'         Torque Control
+                        'C'         Current Control
+
+        action_type:    'AbcCont'   Continuous Action Space in ABC-Coordinates
+                        'Finite'    Discrete Action Space
+    """
+
+    motor_type = 'PMSM'
+    control_type = 'S'
```
atra94, line 22: I think it would be more readable and the structure of the environments naming is displayed better, if the control_type would be changed to SC/TC/CC instead of the single S/T/C at this point and in the other example.

examples/classic_controllers/classic_controllers.py
```diff
@@ -900,8 +1632,8 @@ def reset(self):
     'pid_controller': [ContinuousActionController, ContinuousController, PID_Controller],
```
atra94, line 1632: Also these class names violate the PEP8 conventions. The recommended naming would be PIDController. Please change it also for all remaining classes (PController / PIController / ThreePointController etc.)

examples/classic_controllers/classic_controllers.py
```diff
@@ -632,113 +828,636 @@ def __init__(self, environment, stages, ref_states, color='g', **controller_kwar
...
+    def get_psi_idx_mtpf(self, psi):
+        return np.clip(int((self.psi_count - 1) - round(psi / self.psi_max_mtpf * (self.psi_count - 1))), 0,
+                       self.psi_count)
+
+    def get_t_idx_mtpc(self, torque):
+        return np.clip(int(round((torque + self.max_torque) / (2 * self.max_torque) * (self.t_count - 1))), 0,
+                       self.t_count)
+
+    def control(self, state, torque):
+
```
atra94, line 1310: For this central method, a docstring as well as some inline comments are necessary.

atra94 (commented):
Some minor points are still open that can be closed quickly. However, good work. This will be a good improvement to GEM for the next release.

examples/classic_controllers/classic_controllers.py
```diff
@@ -603,25 +786,38 @@ def control(self, state, reference):
...
+            self.q_controller.reset()
 
 
-class Cascaded_FOC_Controller(Controller):
-    def __init__(self, environment, stages, ref_states, color='g', **controller_kwargs):
+class CascadedFieldOrientedController(Controller):
+    """
+        This controller is used for torque or speed control of synchronous motors. The controller consists of a field
+        oriented controller for current control, an efficiency-optimized torque controller and an optional speed
+        controller. The current control is equivalent to the current control of the FOC_Controller. The torque
```
wkirgsn, line 805: the class' name is not FOC_Controller anymore as I can oversee it

examples/classic_controllers/custom_classic_controllers_synch_motor_example.py
```diff
@@ -0,0 +1,114 @@
...
+    # simulate the environment
+    for i in range(10001):
+        env.render()
+        action = controller.control(state, reference)
+        (state, reference), reward, done, _ = env.step(action)
+        if done:
+            env.reset()
+            controller.reset()
+
+    env.close()
```
wkirgsn, line 114: add new line at end of file

wkirgsn (approved):
all fine now in my opinion, thank you for your effort

max-schenke (approved):
I think it is done for now.
Thank you!

atra94 (approved):

PR #147, opened 2021-07-17
Issue 146

### LINKED ISSUE #146: Warn users on change of syntax
GEM is very forgiving when giving the wrong variables in the wrong places. When changing the syntax of GEM it could happen that old code does not work anymore as intended but does still run without throwing any error or warning.

As an example I have had my problems with the Constraints (former Constraint_Monitor) today. The first syntax we had was to give a key "constraint_monitor" to the reward function. It seems like this is deprecated now as my old constraint monitor was never called. The training, however, did proceed without any complains.

When trying to fix that I looked up the example in "GEM_cookbook.ipynb". The example implements the following Constraint:
```

class MyConstraint(gem.constraints.Constraint)
    
    i_sd_idx = 0
    
    def set_env(self, env):
        self.i_sd_idx = env.physical_system.state_positions['i_sd']
        
    def __call__(self, state):
        """This function returns a "violation degree" within [0,1].
        
        This exemplary constraint returns 0 if i_sd is 0 or below.
        It returns a value between 0 and 1 if i_sd is at most half of the generally allowed current limit.
        If i_sd exceeds 0.5 of the current limit, the violation degree is 1.0 and the episode will be terminated.
        
        The violation degree between 0 and 1 directly affects the reward, but does not end an episode.
        It models an "undesired behavior" the agent shall learn to avoid.
        """
        i_sd = state[self.i_sd_idx]
        return 2 * min(max(i_sd, 0), 0.5)
```

An object of this class is then passed to gem.make() with the constraints key. This example is working such that the __call__ function is indeed called at each step. However, the set_env function is deprecated, too. It has been replaced by set_modules(). The above code runs without throwing any error because i_sd_idx has been set as a class variable.

What I suggest is to have at least a stronger documentation on syntax changes for each update. The most nice-to-have feature would be to catch old syntax in code and throw an error or warning.

atra94:
Deleted the arbitrary keyword arguments in all classes such that now a false kwarg throws an error every time

Closes #146


max-schenke (approved):
All tests run through, the changes are reasonable. 
I will test a little further but for the moment I am happy with this PR. 
Thanks for the effort!

max-schenke (changes requested):
grafik

Unfortunately I found something right away. 
Was this example trying to use a removed feature such that this line can safely be removed from the example or does the error lie somewhere else?

max-schenke (commented):
In `voltage_supplies.py`, I think there are also some more corrections necessary. I was still able to pass unusable keyword arguments and scanning through the file I still find some `**__`.

gym_electric_motor/physical_systems/electric_motors/squirrel_cage_induction_motor.py
```diff
@@ -150,156 +150,3 @@ def _update_initial_limits(self, nominal_new={}, omega=None):
                                zip(self.FLUXES, flux_alphabeta_limits)}
         flux_nominal_limits.update(nominal_new)
         super()._update_initial_limits(flux_nominal_limits)
-
-
-class DoublyFedInductionMotor(InductionMotor):
-    """
```
XyDrKRulof, line 156: Is the removal of the DoublyFedInductionMotor intended here?

XyDrKRulof (changes requested):

gym_electric_motor/physical_systems/electric_motors/squirrel_cage_induction_motor.py
XyDrKRulof: If yes this can be ignored. There is just no mention of it in the commit message, maybe this is a slip up.

XyDrKRulof:
It seems like my old code is now completely broken, just as intended. There were even some syntax changes I did not know about yet (which I hope are currently still nightly only) - e.g. giving a motor dict instead of motor_parameter directly. I will do some scanning on further arbitrary keyword arguments but apart from that I am very happy with this PR.

gym_electric_motor/physical_systems/electric_motors/squirrel_cage_induction_motor.py
atra94: Yes. As you can see it is in the file "squirrel_cage_induction_motor.py". Somehow, it was a duplicate to the motor defined in "doubly_fed_induction_motor.py". Therefore, I removed it.

atra94:
> 
> 
> In `voltage_supplies.py`, I think there are also some more corrections necessary. I was still able to pass unusable keyword arguments and scanning through the file I still find some `**__`.

Yes, thank you. I missed that file.

atra94:
> 
> 
> grafik
> 
> Unfortunately I found something right away.
> Was this example trying to use a removed feature such that this line can safely be removed from the example or does the error lie somewhere else?

No this line has nothing to do with these features.
I fixed this bug and piggybacked it to this Pull Request.

atra94:
> 
> 
> It seems like my old code is now completely broken, just as intended. There were even some syntax changes I did not know about yet (which I hope are currently still nightly only) - e.g. giving a motor dict instead of motor_parameter directly. I will do some scanning on further arbitrary keyword arguments but apart from that I am very happy with this PR.
> 
> edit: Couldn't find anything besides what Max has already discovered in the voltage supplies file

Yes, there are several changes in the user-api waiting in the nightly to be pushed which will make the usage of GEM more user friendly, hopefully. These changes are still in the nightly but will be released soon. 

I could help you to transfer your code to the current nightly version of GEM, if you like to.

XyDrKRulof (approved):
All remarks have been fixed. Thank you

max-schenke:
grafik

I still get an error, now it's a different one. 
Please have another look at it.

atra94:
> 
> 
> grafik
> 
> I still get an error, now it's a different one.
> Please have another look at it.

Fixed

max-schenke:
We found that the motor_parameter dict can still be filled with arbitrary words and decided to catch that as well:
grafik

The corresponding fix will presumably be the last commit within this pull request.

atra94:
Fixed

max-schenke:
Thank you!

One last remark (I hope): I can still provide arbitrary keys in the limits and nominal values' dicts.
Maybe we can simply catch these in the same way and then ship this PR?

grafik

max-schenke (approved):
Perfect. Thank's a lot!

PR #150, opened 2021-07-26
Docs revision

max-schenke:
I applied some changes to the docs (especially the main README).

I have some trouble to locally build the docs for some of the files, e.g., for the reward functions it does not look as intended. Maybe there are differences in the building routine(?).

Also some circuit diagrams still need to be revised. More suggestions are welcome!


atra94 (approved):
Good work. I have got nothing to add. All equations that I have seen render nicely on the HTML pages.

PR #151, opened 2021-08-03
bugfix in the switched_reference_generator.py

atra94:
The SwitchedReferenceGenerator did not define the names of the referenced states within it. (the self._reference_names). 

Additionally, the interface for its creation was simplified. The tests have been adapted.


fbook98 (approved):

max-schenke:
I think at least the file `pmsm_mpc_dq_current_control.py` makes use of the switched reference generator. Please adapt this example (and maybe also other examples) to the proposed changes if necessary.

PR #156, opened 2021-08-08
Finished Env. documentation in sphinx

atra94:
I added the missing environment pages to the Code documentation.


max-schenke:
Thanks again for the effort!

When building the docs I notice that the sideframe is now pretty bloated but that each individual linked page is empty. Can we avoid that somehow? There is no necessity to link each env in the sideframe if there is no content attached.
grafik

max-schenke:
The same way we could maybe just avoid having this list? The links only link to the same empty pages and the motor names are already stated in the table above. (On a side note also the title "DC Motor Environments" is now outdated but if we remove the list entirely this is not really important.)
grafik

atra94:
I am wondering why these pages are empty for you. I get the formatted documentation of each environment 
grafik

However, you are right that the long list bloats everything up. I will try to categorize all environments by its corresponding motor such that the long list will get another hierarchy level.

max-schenke:
My sphinx installation was faulty again. 
The individual env pages are in fact not empty when building the docs correctly!
Nonetheless I think it would look nicer if we could avoid endless lists of envs.

atra94:
grafik
Now the sidebar is updated and looks like this.

max-schenke:
Looks really nice, however I seem to have a buggy interface here:
grafik
Envs main page: looks fine!


grafik
Envs sub page: looks fine!


grafik
Specific motor env page: the categories in the sidebar are gone and I see the full list instead.


As always this could just be a sphinx issue, but maybe you have an idea what causes this?
If not, chances are that the public docs will be built correctly such that this will not happen to users.

atra94:
Sphinx does not always build every page. So try to delete everything in the _build folder. Then, it has to create every page again. 
I will investigate if there is also an option like "make html -new" to tell sphinx to build it again from scratch.

@wkirgsn do you know about such a feature?

max-schenke:
Nice one, deleting and remaking the _build folder was the solution.

No further objections from my side. Thank's for this update!

max-schenke (approved):

PR #157, opened 2021-08-18
Delete tensorforce example

max-schenke:
In order to allow ourselves faster turnaround times concerning GEM itself we decided to discard the tensorforce example for now.
It may come back again and of course motivated users are encouraged to help us with it.


atra94 (changes requested):
You added the "saved agents" folder with multiple binary files.

It would pollute the repository unneccessarily if we delete the files again in the next commit. 

I would suggest to delete the branch and redo the deletion to delete the binary files completely from this repository.

max-schenke:
My bad

PR #161, opened 2021-09-15
Ode solver parameterization

atra94:
The constant static load term of the Polynomial Static Load has been adapted to have a linear behavior in the region near zero angular velocity to stabilize the ODE-solver. 

A linear interpolation between the positive and the negative constant torque seems to be a very stable solution. It is even more stable than the already discussed solution: To use a 0 torque between the positive and negative torque level.

The scipy.ode integrator solves the ODE now very stable, so it has been switched to the default solver.

Furthermore, a little bug has been fixed in the jacobian of the Constant Speed Load

### DIFF
--- a/gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
+++ b/gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
@@ -8,10 +8,32 @@ class PolynomialStaticLoad(MechanicalLoad):
     """ Mechanical system that models the Mechanical-ODE based on a static polynomial load torque.
 
     Parameter dictionary entries:
-        | a: Constant Load Torque coefficient (for modeling static friction)
-        | b: Linear Load Torque coefficient (for modeling sliding friction)
-        | c: Quadratic Load Torque coefficient (for modeling air resistances)
-        | j_load: Moment of inertia of the mechanical system.
+        - :math:`a / Nm`: Constant Load Torque coefficient (for modeling static friction)
+        - :math:`b / (Nm s)`: Linear Load Torque coefficient (for modeling sliding friction)
+        - :math:`c / (Nm s^2)`: Quadratic Load Torque coefficient (for modeling air resistances)
+        - :math:`j_load / (kg m^2)` : Moment of inertia of the mechanical system.
+
+    Usage Example:
+        >>> import gym_electric_motor as gem
+        >>> from gym_electric_motor.physical_systems.mechanical_loads import PolynomialStaticLoad
+        >>>
+        >>> # Create a custom PolynomialStaticLoad instance
+        >>> my_poly_static_load = PolynomialStaticLoad(
+        ...     load_parameter=dict(a=1e-3, b=1e-4, c=0.0, j_load=1e-3),
+        ...     limits=dict(omega=150.0), # rad / s
+        ... )
+        >>>
+        >>> env = gem.make(
+        ...     'Cont-SC-ExtExDc-v0',
+        ...     load=my_poly_static_load
+        ... )
+        >>> done = True
+        >>> for _ in range(1000):
+        >>>     if done:
+        >>>         state, reference = env.reset()
+        >>>     env.render()
+        >>>     (state, reference), reward, done, _ = env.step(env.action_space.sample())
+
     """
 
     _load_parameter = dict(a=0.0, b=0.0, c=0., j_load=1e-5)
@@ -40,9 +62,9 @@ class PolynomialStaticLoad(MechanicalLoad):
     def __init__(self, load_parameter=None, limits=None, load_initializer=None):
         """
         Args:
-            load_parameter(dict(float)): Parameter dictionary.
-            limits(dict):
-            load_initializer(dict):
+            load_parameter(dict(float)): Parameter dictionary. Keys: ``'a', 'b', 'c', 'j_load'``
+            limits(dict): dictionary to update the limits of the load-instance. Keys: ``'omega'``
+            load_initializer(dict): Dictionary to parameterize the initializer.
         """
         load_parameter = load_parameter if load_parameter is not None else dict()
         self._load_parameter = update_parameter_dict(self._load_parameter, load_parameter)
--- a/gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
+++ b/gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
@@ -52,7 +52,7 @@ class PolynomialStaticLoad(MechanicalLoad):
         self._b = self._load_parameter['b']
         self._c = self._load_parameter['c']
 
-    def _static_load(self, omega, *_):
+    def _static_load(self, omega):
         """Calculation of the load torque for a given speed omega."""
         sign = 1 if omega > 0 else -1 if omega < -0 else 0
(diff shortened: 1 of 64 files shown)


gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
```diff
@@ -47,18 +53,26 @@ def __init__(self, load_parameter=None, limits=None, load_initializer=None):
         self._c = self._load_parameter['c']
 
     def _static_load(self, omega, *_):
-        """
-        Calculation of the load torque for a given speed omega.
-        """
-        sign = 1 if omega > 0 else -1 if omega < 0 else 0
-        return sign * (self._c * omega**2 + self._b * abs(omega) + self._a)
+        """Calculation of the load torque for a given speed omega."""
+        sign = 1 if omega > 0 else -1 if omega < -0 else 0
```
max-schenke, line 57: For ease of reading could we also use `sign = np.sign(omega)`?

gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
```diff
@@ -47,18 +53,26 @@ def __init__(self, load_parameter=None, limits=None, load_initializer=None):
...
+            if abs(omega) > self._a / self._j_total * self.tau_decay \
+            else self._j_total / self.tau_decay * omega
+        return sign * self._c * omega**2 + self._b * omega + a
 
     def mechanical_ode(self, t, mechanical_state, torque):
         # Docstring of superclass
-        return np.array([(torque - self._static_load(mechanical_state[self.OMEGA_IDX])) / self._j_total])
+        omega = mechanical_state[self.OMEGA_IDX]
+        static_load = self._static_load(omega)
+        load = torque - static_load
```
max-schenke, line 68: I do not like that we call the sum of torques the "load" here.
I would rather suggest to distinguish "drive_torque", "load_torque" and "total_torque" (or something along those lines).

gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
```diff
@@ -47,18 +53,26 @@ def __init__(self, load_parameter=None, limits=None, load_initializer=None):
...
+        return np.array([load / self._j_total])
 
     def mechanical_jacobian(self, t, mechanical_state, torque):
         # Docstring of superclass
-        sign = 1 if mechanical_state[self.OMEGA_IDX] > 0 else -1 if mechanical_state[self.OMEGA_IDX] < 0 else 0
-        return np.array([[(-self._b * sign - 2 * self._c * mechanical_state[self.OMEGA_IDX])/self._j_total]]), \
-            np.array([1/self._j_total])
+        omega = mechanical_state[self.OMEGA_IDX]
+        sign = 1 if omega > 0 else -1 if omega < 0 else 0
```
max-schenke, line 74: see above

gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
```diff
@@ -47,18 +53,26 @@ def __init__(self, load_parameter=None, limits=None, load_initializer=None):
...
+        # Linear region of the constant load term 'a' ?
+        a = 0 if abs(omega) > self._a * self.tau_decay / self._j_total else self._j_total / self.tau_decay
+        return np.array([[(-self._b * sign - 2 * self._c * omega + a) / self._j_total]]), \
```
max-schenke, line 77: Shouldn't "a" have a negative sign as well?

max-schenke (commented):
First of all thank you for this work!
Could you also report on the achieved speedup using the new solver?

The code looks fine and I will test the functionality next.

Maybe for the future it would be a good idea to move the configuration of the standard solver to the ElectricMotorEnvironment class, which would have spared a lot of file changes in this PR.

max-schenke:
I had no issues running the code, please comment on my remarks and correct the code if necessary.

Also, I am unsure whether any of our examples currently presents the usage of the PolynomialStaticLoad... maybe we should add that in some place. (Or did I just miss it?)

gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
atra94: Yes it would be possible. However, it has shown that np.sign is quite slow on scalars compared to this if-else clause. And speed is important within the ODE methods.

grafik

gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
atra94: No, the derivative of the load ODE over omega is calculated at those lines. In the old version, the constant torque "a" has been completely eliminated by the derivation (except for a non-modeled infinite slope at omega=0). 
Now, there is a non-infinite slope between the two levels of "a" for negative and positive speeds. 
Therefore, the derivative of 'a' at that point is constant `self._j_total / self._tau_decay (>0)`.

gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
atra94: A good point, thank you. 
Solely, I will keep the name of the input as torque because is inherited by the base-class "mechanical load". Furthermore in the documentation the input torque is documented correctly.

atra94:
> 
> 
> I had no issues running the code, please comment on my remarks and correct the code if necessary.
> 
> Also, I am unsure whether any of our examples currently presents the usage of the PolynomialStaticLoad... maybe we should add that in some place. (Or did I just miss it?)

[Here](https://github.com/upb-lea/gym-electric-motor/blob/master/examples/classic_controllers/classic_controllers_dc_motor_example.py) is an example using a speed control environment. That ones use the PolynomialStaticLoad.

atra94:
> 
> 
> First of all thank you for this work!
> Could you also report on the achieved speedup using the new solver?
> 
> The code looks fine and I will test the functionality next.
> 
> Maybe for the future it would be a good idea to move the configuration of the standard solver to the ElectricMotorEnvironment class, which would have spared a lot of file changes in this PR.

**Setting 1:**
 - Default ExtExDc Speed Control Environment
 - Classic PI Controller used
 - No visualization
 - 10k steps
 
- Scipy.ode solver : 5.05s
- Euler Solver :  3.58s
- Scipy.solve_ivp Solver : 8.26s
- Scipy. Odeint solver : 9.23s

**Setting 2:**
 - Default PMSM Current Control Environment
 - Classic PI Controller used
 - No visualization
 - 10k steps

 
Scipy.ode solver : 4.59s
Euler Solver :  3.26s
Scipy.solve_ivp Solver : 7.90s
Scipy. Odeint solver : 8.24s

Yes, it could be benefitial to use this as standard parameter. However, all environments are currently configured completely in their respective classes. You have always a good overview of what the environment consists then. I am also not completely satisfied with the definition of our environments currently (Way too many files and classes).

gym_electric_motor/physical_systems/mechanical_loads/polynomial_static_load.py
max-schenke: Good point, I didn't expect this to be the faster approach.

max-schenke (approved):
Thank you very much!

PR #163, opened 2021-09-23
Classic controllers update

fbook98:


examples/classic_controllers/classic_controllers.py
```diff
@@ -1471,6 +1598,595 @@ def reset(self):
...
+    def get_plot_data(self):
+        return dict(ref_state=[], ref_value=[], external=[[self.psi_abs], [self.psi_angle]])
+
+    def reset(self):
+        self.flux_observer.reset()
+        self.d_controller.reset()
+        self.q_controller.reset()
+
+
+class CascadedFieldOrientedControllerRotorFluxObserver(Controller):
```
atra94, line 1708: I think, the naming of all new classes qith the postfix "RotorFluxObserver" is a little imprecise. This is a controller and not an observer.

Maybe it could be named `InductionMotorFieldOrientedController`  or even `RotorFluxObserverFieldOrientedController` @max-schenke what do you think about it?

atra94 (changes requested):
Thank you for the good work.
I have multiple issues but they are all minor problems.

First of all, I would like to have more code documentation for better readability. This is missing at many places and makes the code hard to understand. Especially, every non-inherited method should have a docstring and long or complex methods should also have inline commenting. 

Furthermore, much of the new code is copied and pasted from anywhere else with little adjustments  (mostly from the classic FOC). If there are really some easy to reuse lines and methods, I would appreciate if those could be reused through inheritance, for example.
But this is only a nice to have feature.

Finally, the classic_controller.py file consists of 2300 lines now. I think it should be split up into multiple files (one per class) for a better structure and readability.

After you commented the code, I will dive into the details and test it.

gym_electric_motor/visualization/motor_dashboard.py
```diff
@@ -272,3 +272,7 @@ def _update(self):
         for fig in self._figures:
             fig.canvas.draw()
             fig.canvas.flush_events()
+
+    @property
```
atra94, line 276: Please document this property. Furthermore, it is a common practice to put the properties on top of a class definition (after the docstring)

atra94 (commented):
There are still minor changes required that have not been solved.

examples/classic_controllers/controllers/induction_motor_cascaded_foc.py
```diff
@@ -0,0 +1,203 @@
...
+        self.nominal_values = self.env.physical_system.nominal_state
+        self.tau_sigma = self.sigma * self.l_s / (self.r_s + self.r_r * self.l_m ** 2 / self.l_r ** 2)
+        self.tau = self.env.physical_system.tau
+
+        self.dq_to_abc_transformation = environment.physical_system.dq_to_abc_space
+
+        self.torque_control = 'torque' in ref_states or 'omega' in ref_states
+        self.current_control = 'i_sd' in ref_states
+        self.omega_control = 'omega' in ref_states
+        self.ref_state_idx = [self.i_sq_idx, self.i_sd_idx]
```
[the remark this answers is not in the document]
atra94: @max-schenke This comment was regarding the order of the currents, right? Then, it is not fixed yet.

atra94 (approved):
I think there is one open point from max. The rest is fine now. 👍

Issue #116, opened 2020-10-31
Investigation of floating type precision


atra94:
When running the gym-environments, currently gym throws the following warning regarding the precision of the observation and action space:
![grafik](https://user-images.githubusercontent.com/50244395/97779609-7143d700-1b7f-11eb-97ab-95c4f359ee91.png)

I have looked for the origin of this warning. It is generated, when the "Box" spaces for action and observation are instantiated. 

Gym expects as default a np.float32 whereas the default of numpy is np.float64. If one default is overwritten during the creation of the spaces, the warning will disappear.

Maybe it would also be useful to fix the precision throughout the environment and to make it selectable by the user. A lower precision could speed up the calculation at the cost of accuracy.

PR #168, opened 2021-10-22
Fixed issue 116. Warning on floating type precision

atra94:
By adding the explicit datatype `np.float64` to each `gym.spaces.Box` instantiation, the warning that has showed up at every environment creation (c.f. #116) is avoided.

I also investigated, if a change to `np.float32` could speed up the calculations. However, it seems to have no effect for typical array length (~10) in GEM. 

(see image)
In the upper case, a calculation with arrays of dtype np.float64 is executed and in the lower case with np.float32. 

grafik


max-schenke (approved):
No objections. Thanks for the effort!
We need to keep an eye on that warning in case we missed to change to float64 in some places.
However, I see no scenario where this could cause any errors. Therefore I will approve right away.

max-schenke:
Please address the merge conflict and then we can finalize this PR.

PR #173, opened 2021-11-29
Improved SC Env Parameterization Closes #137

### LINKED ISSUE #137: Standard configuration for speed reference generator is too dynamic
During testing of the classic controller updates I have observed that the reference generator signal for the speed control case is always very dynamic, cf. the following example with a series DC motor:
Unbenannt

Compared to the current/torque references, which can be as dynamic as that, the speed control loop is not able to follow such high dynamic reference profiles due to the bandwidth limitations of the inner control loops and (more importantly) due to the rotor / load inertia. For realistic speed reference profiles the standard speed profile generator should be at least by one order of magnitude less dynamic than currently configured.

atra94:
Improved parameterization of the references for the speed control environments. 

Tested with gem_control controllers


max-schenke (approved):
No objections. Thank you.

PR #179, opened 2022-01-11
classic controller custom examples bugfix

fbook98:
Bugfix


max-schenke:
Thanks a lot for this fix!

PR #181, opened 2022-01-25
Ipympl compatibility

atra94:
Solved the compatibility problem with the %matplotlib widget backend and vs code.

If the environment is executed in a notebook, it is now recommended to call ``visu.initialize()`` in another cell as the execution and the ``env.render()`` calls. The plots will be shown below the cell with the initialize-call.

Furthermore: Added a note for missing Colab compatibility


bhk11:
Functionality is identical for me in PyCharm and VSCode. I could not find any difference in the behavior to the master in PyCharm.

examples/model_predictive_controllers/pmsm_mpc_dq_current_control.ipynb
```diff
@@ -1110,9 +343,9 @@
    ],
    "source": [
     "%matplotlib notebook\n",
-    "\n",
+    "visu = MotorDashboard(state_plots=['i_sq', 'i_sd', 'u_sq', 'u_sd'], update_interval=10), # visualization\n",
     "env = gem.make('AbcCont-CC-PMSM-v0',\n",
-    "            visualization=MotorDashboard(state_plots=['i_sq', 'i_sd', 'u_sq', 'u_sd'], update_interval=10), # visualization\n",
+    "            visualization=visu\n",
```
bhk11, line 348: visu,

examples/model_predictive_controllers/pmsm_mpc_dq_current_control.ipynb
```diff
@@ -1110,9 +343,9 @@
    ],
    "source": [
     "%matplotlib notebook\n",
-    "\n",
+    "visu = MotorDashboard(state_plots=['i_sq', 'i_sd', 'u_sq', 'u_sd'], update_interval=10), # visualization\n",
```
bhk11, line 346: image
The comma in the end of the line make visu to a tuple in VSCode. Without comma in woks fine to initialize visu.

examples/model_predictive_controllers/pmsm_mpc_dq_current_control.ipynb
bhk11: line 360: u_sup has to be changed to u_nominal for the actual version of gem.

bhk11: Info: When I start the last two cells again, the reference generators don't work correct. I guess, the generators are using a default value in this case. This is probably a new issue/enhancement.

image

examples/reinforcement_learning_controllers/stable_baselines3_dqn_disc_pmsm_example.ipynb
```diff
@@ -411,6 +412,15 @@
     "Once you've setup the environment and defined your parameters starting the training is nothing more than a one-liner. For each algorithm all you have to do is call its ```learn()``` function. However, you should note that the execution of the training can take a long time. Currently, Stable-Baselines3 does not provide any means of saving the training reward for later visualization. Therefore, a ```RewardLogger``` callback is used for this environment (see code a few cells above)."
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "motor_dashboard.initialize()"
```
bhk11: The jupyter kernel dies after successful execution of the cell.

image

I tried to execute the example in PyCharm. Also no success and no error message.

bhk11 (changes requested):
Added some comments with changes.

I wasn't able to find the problem in the stable baseline example. When I execute the cell with the initialization routine for the motor dashbord, the python kernel dies in VSCode and no termination of the cell in PyCharm.

examples/model_predictive_controllers/pmsm_mpc_dq_current_control.ipynb
atra94: That is no error, just a StepReference Generator. This environment switches the reference generator every n-steps between a sinusoidal, triangular and step reference.

examples/model_predictive_controllers/pmsm_mpc_dq_current_control.ipynb
atra94: Thanks, that was my bad.

examples/reinforcement_learning_controllers/stable_baselines3_dqn_disc_pmsm_example.ipynb
atra94: Interesting. On my machine it works fine. Not only in the browser but also in VSCode. However, I just saw that the training of the agent runs without rendering. Therefore, this cell was deleted.

examples/model_predictive_controllers/pmsm_mpc_dq_current_control.ipynb
atra94: Yes, under these circumstances, I can reproduce the error. I was also able to identify the source: A new env. is created but the reference generators are not instantiated again. So, you use the same instances as before. Somehow, this fails. 
If you run also the third to last cell and create new ref gens, everything works fine. 

At first sight, I was not able to locate the error. However, I will fix this soon within another PR (as you already mentioned)

bhk11 (approved):
Looks good, thanks.

examples/reinforcement_learning_controllers/stable_baselines3_dqn_disc_pmsm_example.ipynb
bhk11: This is a problem of my environment, because some package versions are incompatible with stable basline. If I want to use stable baseline, I have to set up a new environment.

PR #187, opened 2022-02-25
Tau converter fix

atra94:
Fixed the issue concerning different values of tau in the converter and the system.


max-schenke:
Very nice.
Could you also have a look at the external_speed_load.py ?
I think we have the same issue there, maybe you can find a solution for that as well.
As always, many thanks!

max-schenke (approved):
Good call, then we can merge this.

PR #205, opened 2022-12-20
improve doc on difference between RMS and peak-to-peak voltage

wkirgsn:
as well as phase vs. line voltage


wkirgsn:
merge PR despite errors since these are due to version mismatches in Python and numpy and need attention elsewhere.
This PR just updates documentation

PR #237, opened 2024-04-25
Sphinx fix

XyDrKRulof:
The Sphinx version pre-installed in the GitHub action was too old for the Sphinx extensions we use. This pull request should fix this by forcing an update for the sphinx version inside the action


bhk11 (approved):
After successfully creating the doc, it looks good to me.

PR #241, opened 2024-05-20
Dont raise exception on dict input

devandt:
Fixes problems described in the email by Darius

### DIFF
--- a/src/gym_electric_motor/utils.py
+++ b/src/gym_electric_motor/utils.py
@@ -2,7 +2,6 @@ import gymnasium
 import numpy as np
 
 
-# This hacky function stays for now because all envs uses it and refectoring would be a pain
 def initialize(base_class, arg, default_class, default_args):
     if arg is None:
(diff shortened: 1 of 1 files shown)


src/gym_electric_motor/utils.py
```diff
@@ -2,18 +2,17 @@
 import numpy as np
 
 
-# This hacky function stays for now because, all envs use it and refectoring would be a pain TODO: fix this
+# This hacky function stays for now because all envs uses it and refectoring would be a pain
```
XyDrKRulof, line 5: Can we elaborate on this more? Either remove this comment and raise an issue explaining why this function should be removed and how it is difficult to do so. Otherwise, please make this comment more elaborate (also make it sound more neutral). Thank you

XyDrKRulof (changes requested):
Can be merged once the comment is fixed/removed

XyDrKRulof (approved):

Issue #242, opened 2024-05-28
Allow EESM parameters that are already transformed to stator side


max-schenke:
In the current configuration, GEM expects EESM parameters to be the physical quantities of either the rotor or the stator side. The rotor-side parameters are then by default transformed to the stator side for further calculations.

For convenience reasons, it should be allowed to specify rotor parameters that are already transformed to the stator side. The variables have different names and the transformation would need to be skipped in that case (or reverted to yield the physical circuit quantities, which are not needed for calculation but could then be retrieved from the environment).

PR #243, opened 2024-05-28
EESM rotor parameters can be defined untransformed or transformed to stator side

max-schenke:
Related to #242 

So far I had trouble to test this one and would need assistance.


max-schenke:
Getting error message:

KeyError: 'Cannot update_dict the source_dict. The key "l_m" is not available.'

I guess GEM checks whether all keys from the default case are available in the user-specified case. I specified "l_m" whereas the default case specifies "l_M", and I think this mismatch causes the error message.

PR #244, opened 2024-05-29
Fix to Python 3.8 being deprecated for Sphinx-Build

XyDrKRulof:
As the old GitHub action relies on Python 3.8 which is not supported by GEM anymore, we will have to explore (temporary) solutions for this.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -31,15 +31,12 @@ jobs:
     - name: Test with pytest
       env:
         PY_VER: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


XyDrKRulof:
I was able to find a repository which implements GitHub actions with the degree of freedom to chose the underlying Sphinx Docker container version. I chose 7.0.0 for now. Documentation seems to be back. In the future we might want to remove 3rd party repositories from our GitHub actions and implement the build routine ourselves.

bhk11 (approved):
docs build is successful.

Issue #225, opened 2023-12-22
Extend the GEM cookbook to explain voltage supply usage


XyDrKRulof:
The GEM cookbook features a small paragraph explaining the usage of voltage supply and then an examplary gem.make() where the voltage is supply is being omitted:

```
# define a PMSM with discrete action space
env = gem.make(  
    motor_env_id,
    # visualize the results
    visualization=visualization,
    # parameterize the PMSM and update limitations
    motor=dict(
        motor_parameter=motor_parameter,
        limit_values=limit_values,
        nominal_values=nominal_values,
        motor_initializer=motor_initializer,
    ),
    # define the random initialisation for load and motor
    load=load,
    reward_function=rf,
    tau=tau,
    #supply=supply,
    reference_generator=rg,
    ode_solver='euler',
    callbacks=my_callback,
    physical_system_wrappers=physical_system_wrappers, # Pass the Physical System Wrappers
    constraints=constraints
)```

However, it is extremely important to set the voltage supply to the limit voltage in case of a constant voltage supply. If you don't, the motor will behave very unexpected.

Another approach could be to automatically set the supply voltage to the limit voltage if nothing else is specified. Or to throw an error that the supply voltage was not given by the user while nominal and limit values were.

PR #256, opened 2024-11-14
Updated the GEM_cookbook to work with GEM 3.0.0

XyDrKRulof:
And fix issue #225. Additionally, the deprecation error message for the initialization in utils.py was changed from 2.0 to 3.0.0


bhk11 (approved):
changes looks complete

PR #261, opened 2024-11-20
test for classic controllers, dc motor, dcExternallyExcited motor

### LINKED ISSUE #254: Add further automated testing of existing features

ranil345:
Could you please review these changes
1. ...\gym-electric-motor\tests\test_controllers_examples\test_classic_controllers.py
2. ..\gym-electric-motor\tests\test_physical_systems\test_electric_motors.py


tests/test_physical_systems/test_electric_motors.py
```diff
@@ -0,0 +1,253 @@
...
+    DcExternallyExcitedMotor,
+    DcPermanentlyExcitedMotor
+)
+from gymnasium.spaces import Box
+import numpy as np
+
+#parameters for dc motor
+test_dcmotor_parameter = {"r_a": 16e-2, "r_e": 16e-1, "l_a": 19e-5, "l_e_prime": 1.7e-2, "l_e": 5.4e-1, "j_rotor": 0.025}
+test_dcmotor_nominal_values = dict(omega=350, torque=16.5, i=100, i_a=96, i_e=87, u=65, u_a=65, u_e=61)
+test_dcmotor_limits = dict(omega=350, torque=38.0, i=210, i_a=210, i_e=215, u=60, u_a=60, u_e=65)
```
bhk11: I'm not sure if it's good to set the limits lower then the nominal values for the voltages. (parameters starting with an u)

ranil345: @bhk11 Sure, I will set try to update all the limits values to be higher than nominal values

examples/classic_controllers/controllers/cascaded_foc_controller.py
```diff
@@ -28,25 +28,25 @@ def __init__(
         torque_control="interpolate",
         **controller_kwargs,
     ):
-        t32 = environment.physical_system.electrical_motor.t_32
-        q = environment.physical_system.electrical_motor.q
+        t32 = environment.unwrapped.physical_system.electrical_motor.t_32
```
bhk11: Nice work, that the test run trough successful!
Personally, I would prefer to unify the code and use `get_wrapper_attr('physical_system')` instead of `unwrapped.physical_system` .
In my opinion this getter should also work for the action_space, state_names and so on.

tests/test_physical_systems/test_electric_motors.py
```diff
@@ -0,0 +1,253 @@
...
+     :param defaultDCMotor:
+     :param concreteDCMotor:
+     :return:
+     """
+     assert defaultDCMotor.motor_parameter ==  {"r_a": 16e-3, "r_e": 16e-2, "l_a": 19e-6, "l_e_prime": 1.7e-3, "l_e": 5.4e-3, "j_rotor": 0.0025}
+     assert defaultDCMotor.nominal_values == dict(omega=300, torque=16.0, i=97, i_a=97, i_e=97, u=60, u_a=60, u_e=60)
+     assert concreteDCMotor.motor_parameter == test_dcmotor_parameter
+     assert concreteDCMotor.nominal_values == test_dcmotor_nominal_values
+     assert concreteDCMotor.limits == test_dcmotor_limits
+     assert concreteDCMotor._initial_states == {"i_a": 10.0, "i_e": 15.0}
```
XyDrKRulof: test_dcmotor_default_initializer["states"] would be better instead of {"i_a": 10.0, "i_e": 15.0}

tests/test_physical_systems/test_electric_motors.py
```diff
@@ -0,0 +1,253 @@
...
+          "i":2,
+          "i_a":3,
+          "i_e": 4,
+          "u_a": 5,
+          "u_e": 6,
+    }
+     
+    extex_state_space = Box(low=-1, high=1, shape=(7,), dtype=np.float64)
+    assert concreteDCMotor._initial_states == default_initial_state
+    concreteDCMotor._initial_states = {'i_a': 100.0, 'i_e': 20.0}
```
XyDrKRulof: concreteDCMotor._initial_states = new_initial_state

tests/test_physical_systems/test_electric_motors.py
```diff
@@ -0,0 +1,253 @@
...
+
+    assert concreteDCMotor._initial_states == new_initial_state
+    assert np.array_equal(concreteDCMotor.reset(extex_state_space,extex_state_positions),default_Initial_state_array)
+    assert concreteDCMotor._initial_states ==  default_initial_state
```
XyDrKRulof: Could you also test the behavior when concreteDCMotor._initializer["states"] is overwritten? In that case we assume the new initial values will not be overwritten on reset.

XyDrKRulof (commented):
Good test structure already! Would be nice to test a few more combinations of test motors with different parameters and different states + state transitions. E.g., initializing a motor in different initial states, applying a voltage and asserting the correct resulting state. Repeat this for different motor parameters. 

For readibility these different parameters shouldn't each have their own fixture, though. Either create an array of Motors in a single fixture or initialize the motors inside the test case. Thanks for the effort :)

examples/classic_controllers/controllers/cascaded_foc_controller.py
ranil345: Hi,
This proposed change has been made and will be pushed with the commit message ""change to get_wrapper_attr()""

tests/test_physical_systems/test_electric_motors.py
ranil345: This is what I have already tried to do with "test_ConcreteDCMotor_reset", first the motor is initialized with a state different from the one in DCMotor.py and then the state is again changed with the "new_initial_state". On reset the intial_State is changed to default_initial_state which is different from the value in the DCMotor.py. I hope this is what you expect. Thank you!

tests/test_physical_systems/test_electric_motors.py
ranil345: Sure, Will make the change

bhk11 (approved):
Added tests and further changes looks fine.
Added the information into the changelog

PR #267, opened 2025-02-19
six phase motor - inprogress

ranil345:
Development of six phase machines - new feature


ranil345:
[sixphasePMSM_ode (1).pdf](https://github.com/user-attachments/files/18987591/sixphasePMSM_ode.1.pdf)
These are the set of equation i have used for the ode, please let me know in case of anything incorrect

src/gym_electric_motor/physical_systems/electric_motors/SixPhase_PMSM.py
```diff
@@ -0,0 +1,169 @@
...
+    ======== ===========================================================
+        
+    """
+    I_SD_IDX = 0
+    I_SQ_IDX = 1
+    I_SX_IDX = 2
+    I_SY_IDX = 3
+    CURRENTS_IDX = [0, 1, 2, 3]
+    CURRENTS = ["i_sd", "i_sq", "i_sx", "i_sy"]
+    VOLTAGES = ["u_sd", "u_sq", "u_sx", "u_sx"]
```
max-schenke: Typo: last voltage component should be u_sy

src/gym_electric_motor/physical_systems/electric_motors/SixPhase_PMSM.py
```diff
@@ -0,0 +1,169 @@
...
+                    state[self.I_SQ_IDX],
+                    omega * state[self.I_SD_IDX],
+                    omega,
+                    u_dqxy[1],
+                    state[self.I_SX_IDX],
+                    omega * state[self.I_SY_IDX],
+                    u_dqxy[2],
+                    state[self.I_SY_IDX],
+                    omega * state[self.I_SX_IDX],
+                    u_dqxy[3]
```
max-schenke: Order of these regressors is rather unsorted. Please implement the same ordering as in other motors, i.e. group the current components, the voltage components and the current * omega components. 
The same reordering must then be applied to the parameter matrix (self._model_constants).

src/gym_electric_motor/physical_systems/electric_motors/SixPhase_PMSM.py
```diff
@@ -0,0 +1,169 @@
...
+        self._update_model()
+        self._update_limits()
+    
+    def _update_model(self):
+        """Updates the motor's model parameters with the motor parameters.
+
+        Called internally when the motor parameters are changed or the motor is initialized.
+        """
+        mp = self._motor_parameter
+        self._model_constants = np.array([
```
max-schenke: It would be helpful to specify via comment which matrix column belongs to which regressor (please see other motor types for reference).

src/gym_electric_motor/physical_systems/electric_motors/six_phase_motor.py
```diff
@@ -0,0 +1,77 @@
...
+            The converted quantities in the dq representation like ''[i_sd, i_sq, i_sx, i_sy, i_s0+, i_s0-]''.
+            since 2N topology is considered (case where the neutral points are not connected) i_s0+, i_s0- will not be taken into account 
+        """
+        t_vsd = 1/ 3 * np.array([
+            [1, -0.5, -0.5, 0.5 * np.sqrt(3), -0.5 * np.sqrt(3), 0],
+            [0, 0.5 * np.sqrt(3), -0.5 * np.sqrt(3), 0.5, 0.5, -1],
+            [1, -0.5, -0.5, -0.5 * np.sqrt(3), 0.5 * np.sqrt(3), 0],
+            [0, -0.5 * np.sqrt(3), 0.5 * np.sqrt(3), 0.5, 0.5, -1],
+            [1, 1, 1, 0, 0, 0],
+            [0, 0, 0, 1, 1, 1]
```
max-schenke: For me this seems inconsistent: the function doc says that i_s0 components will not be considered, but the subsequent matrix computation does still compute them / contains the corresponding lines. I think it should be possible to:
1) either get rid of these lines if the definition is to only hold for the 2N topology,
OR
2) cover the general case, whereas it might as well be desired to simulate a 1N topology.

src/gym_electric_motor/physical_systems/electric_motors/six_phase_motor.py
```diff
@@ -0,0 +1,77 @@
...
+        cos = math.cos(epsilon)
+        sin = math.sin(epsilon)
+        tp_alphaBetaXY = np.array([
+                 [cos, sin, 0, 0, 0, 0],
+                 [-sin, cos, 0, 0, 0, 0],
+                 [0, 0, cos, sin, 0, 0],
+                 [0, 0, -sin, cos, 0, 0],
+                 [0, 0, 0, 0, 1, 0]
+                 [0, 0, 0, 0, 0, 1]
+        ])
```
max-schenke: The first diagonal block entry has a positive sense of rotation, but the second one should have a negative. (As of now, both are coded the same).
For easier comprehension, I would suggest to define the 2D rotation matrix first as a function T(epsilon), and then compound this 6D rotation matrix from blkdiag[T(epsilon), T(-epsilon), eye(2)].

max-schenke (changes requested):
Makes a good first impression!
Please reorder the ODE to make it more easily comprehensible.
I guess we would need to decide which level of generality we / you would like to cover with this model.

src/gym_electric_motor/physical_systems/electric_motors/SixPhase_PMSM.py
```diff
@@ -0,0 +1,187 @@
...
+    i_sc2           A
+
+    =============== ====== =============================================
+    =============== ====== =============================================
+    Motor Voltages  Unit   Description
+    =============== ====== =============================================
+    u_sd            V      Direct axis voltage
+    u_sq            V      Quadrature axis voltage
+    u_sx            V
+    u_sx            V
```
max-schenke: u_sy

src/gym_electric_motor/physical_systems/electric_motors/SixPhase_PMSM.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Called internally when the motor parameters are changed or the motor is initialized.
+        """
+        mp = self._motor_parameter
+        self._model_constants = np.array([
+            #        omega,         i_d,         i_q,         i_x,        i_y, u_d, u_q, u_x, u_y, omega * i_d, omega * i_q, omega * i_x,  omega * i_y
+            [            0,   -mp['r_s'],          0,           0,          0,   1,   0,   0,   0,           0,   mp['l_q'],           0,          0],
+            [-mp['psi_PM'],            0,  -mp['r_s'],          0,          0,   0,   1,   0,   0,   -mp['l_d'],          0,           0,          0],
+            [            0,            0,           0, -mp['r_s'],          0,   0,   0,   1,   0,            0,          0,           0, -mp['l_y']],
+            [            0,            0,           0,          0, -mp['r_s'],   0,   0,   0,   1,            0,          0,   mp['l_x'],          0]
```
max-schenke: Looks very good to me. Good job!

src/gym_electric_motor/physical_systems/electric_motors/six_phase_motor.py
max-schenke: The sense of rotation is still documented in the wrong way here

src/gym_electric_motor/physical_systems/electric_motors/six_phase_motor.py
```diff
@@ -0,0 +1,99 @@
...
+        [math.cos(theta), math.sin(theta)],
+        [-math.sin(theta), math.cos(theta)]
+        ])
+        
+        t1 = rotation_matrix(epsilon)
+        t2 = rotation_matrix(-epsilon)
+        tp_alphaBetaXY = np.block([
+            [t1, np.zeros(t1.shape[0],t1.shape[1])],
+            [np.zeros(t2.shape[0],t2.shape[1]), t2],
+        ])
```
max-schenke: scipy provides a command that should make this snippet more clear

https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.block_diag.html

max-schenke (commented):
Thanks again for your invested time, this is looking very nice already. There are only a few miscellaneous comments I would like you to consider.

ranil345:
I have used these calculations to write down the Jacobian, Review on this would be very helpful
[jacobian.pdf](https://github.com/user-attachments/files/19362100/jacobian.pdf).  Also by the last commit we can now make six phase environment. Now I would like to have your expertise to check whether the environment created really do have the expected features ?
I have also worked on the requested changes . Thank you!

bhk11 (approved):
Looks good to me

docs/parts/physical_systems/electric_motors/sixphase_pmsm.rst
```diff
@@ -0,0 +1,26 @@
+Six Phase Permanent Magnet Synchronous Motor
+##################################
+
+
+Electrical ODE
+**************
+
+.. math::
+    \frac{\mathrm{d} i_s^d}{\mathrm{d} t} = \frac{u_s^d + \omega_\mathrm{el} L_s^q i_s^q - R_s i_s^d}{L_s^d} \\
```
XyDrKRulof: All instances of d,q,x,y,s have to be non-italic (\mathrm{...})

Also: i_s^d = i_d, L_s^d = L_d, L_s^y = L_y ...

XyDrKRulof (changes requested):
Looking great, thanks for all of your efforts :) My last request is to change the doc as noted in my comment

docs/parts/physical_systems/electric_motors/sixphase_pmsm.rst
ranil345: Done, Please look into this commit -  https://github.com/upb-lea/gym-electric-motor/pull/267/commits/74e04d5a3f5cfaa440da15650cd77e1277279902

XyDrKRulof (approved):
Everything seems good to me now. Thanks!

src/gym_electric_motor/physical_systems/electric_motors/six_phase_PMSM.py
```diff
@@ -0,0 +1,262 @@
...
+    =====================  ==========  ============= ===========================================
+    r_s                    Ohm         64.3e-3       Stator resistance
+    l_d                    H           125e-6        Direct axis inductance
+    l_q                    H           126e-6        Quadrature axis inductance
+    l_x                    H           39e-6         x-axis inductance
+    l_y                    H           35e-6         y-axis inductance
+    p                      1            5            Pole pair number
+    psi_PM                 Vs          4.7e-3        flux linkage of the permanent magnets
+    =====================  ==========  ============= ===========================================
+
```
bhk11, line 21: Small comment:
We should add the information here, where we have these parameters from:

The default parameters originate from:
L. Broghammer et al., "Reinforcement Learning Control of Six-Phase Permanent Magnet Synchronous Machines," 2023 13th International Electric Drives Production Conference (EDPC), Regensburg, Germany, 2023, pp. 1-8, doi: 10.1109/EDPC60603.2023.10372153.

PR #282, opened 2025-06-11
274 add some automated model predictive current control to gem control

manjeetjha070:
I have renamed the curred mpc example file to "pmsm_ccs_mpc_dq_current_control.ipynb" and created a new file for finite control set as "pmsm_fcs_mpc_dq_current_control.ipynb".


examples/model_predictive_controllers/pmsm_ccs_mpc_dq_current_control.ipynb
XyDrKRulof, line 1: Would be nice to change the title of this file now to make clear that it is the CCS example. Thanks!

examples/model_predictive_controllers/pmsm_fcs_mpc_dq_current_control.ipynb
XyDrKRulof, line 1: Very nice work already, thank you! :) Some comments:

1. Change the introductionary text and title to reflect that this is an FCS example
2. Remove Gekko (and other unused libraries) - all mentionings, import and installation
3. Implement more than one step prediction for MPC. Utilize self.ph_ for this
4. Create an FCS environment in the gem.make comment. Currently, it is still CCS

Thanks for your effort and in case of questions feel free to reach out for us

XyDrKRulof (changes requested):
Already very nice work. For improvements, see my comments.

manjeetjha070:
I have implemented the following changes plz. review.

1. Changed the introductionary text and title to reflect that this is an FCS example and CCS example.
2. Removed Gekko (and other unused libraries).
3. Implemented generalized MPC. 

Pending point,

1. To create a FCS environment, I would need a guidance for it.

XyDrKRulof (changes requested):
Hey Manjeet,

thanks for your good work - the FCS - MPC algorithm with prediction horizon > 1 is already looking really good :) Here are some further comments that we discussed:

1. Replace the "max_depth" variable in the simulate_sequence function with self.ph_ (which, for readability, you could also rename to self.prediction_horizon)
2. The self.voltage_vectors variable seems to be a mix out of abc and alphabeta coordinates - please change this table to the one you get directly out of the environment:
`self.subactions = -np.power(-1,env.get_wrapper_attr('physical_system')._converter._subactions)`
3. Instead of saving the abc voltage sequence of the lowest cost prediction, save (and return) the indices of them as provided in 2. (e.g., if the best voltage vector is (-1,1,-1) and it is in position 3 return 3)
4. Update epsilon_el for the future predictions and average it for the abc -> dq conversion:
` v_dq = self._forward_transformation(np.array([v_a, v_b, v_c]), epsilon_el + 0.5*omega*tau)`
`epsilon_next = epsilon + omega*tau`

If you have any questions regarding any of these points feel free to contact us! :)

manjeetjha070:
Refactored FCS-MPC implementation to align with environment-provided utilities and improve prediction logic

- Replaced manual abc → αβ → dq voltage transformations with environment's native method: `abc_to_dq_space`
- Updated voltage vector table to use finite set directly from the environment via:
  `self.subactions = -np.power(-1, env.get_wrapper_attr('physical_system')._converter._subactions)`
- Now storing and returning subaction **indices** instead of full abc voltage vectors for better integration and comparison
- Updated epsilon handling in prediction loop:
  - Epsilon is now incremented as `epsilon_next = epsilon + omega * tau`
  - dq-transformation uses averaged value: `epsilon + 0.5 * omega * tau`
- Replaced hardcoded max_depth with `self.prediction_horizon` (renamed from `self.ph_` for clarity)

Previously, I was doing manual calculations: converting abc voltages to αβ and then dq using custom logic.
Now all transformations use the environment's own coordinate conversion function and finite action space, making the controller more accurate and modular.

Thanks again for the helpful feedback!

manjeetjha070:
I have revised the description for FCS-MPC, now its ready to get merge.

manjeetjha070:
Now, the MPC controller calculates the motor-specific state vector and electrical jacobian once in the control method, then uses these for prediction in _simulate_sequence. This makes the controller compatible with different motor types and avoids redundant calculations during the prediction loop. Unsupported motor types will raise a clear error.

bhk11 (approved):
Works fine for me in the mentioned limits in the changelog

PR #286, opened 2025-09-18
277 gem control change the eesm torque control

### LINKED ISSUE #277: GEM-Control: Change the EESM torque control
The current calculation of the current references in the torque control scenario of GEM-Control is not working correct.
Change the optimization routine to maximum torque per copper losses.

Image

bhk11:
Created a draft pull request to check the changes and may find the problem.


src/gem_controllers/stages/operation_point_selection/eesm_ops.py
```diff
@@ -130,139 +132,318 @@ def solve_analytical(self, torque, psi, i_e):
...
+            seeds = [x0, np.array([-0.3*i_s_lim, np.sign(Tref)*0.8*i_s_lim, 0.5*i_f_lim])]
+            for s0 in seeds:
+                try:
+                    res = minimize(obj_loss, s0, method='SLSQP',
+                                   bounds=bnds, constraints=cons_for_T(Tref),
+                                   options=dict(maxiter=200, ftol=1e-9, disp=False))
+                    if res.success and feasible(*res.x):
+                        return res.x
+                except Exception:
+                    pass
```
max-schenke: Looks like this optimization is ran twice per T_ref? Once seeded with x0 and once with some hardcoded currents?

src/gem_controllers/stages/operation_point_selection/eesm_ops.py
```diff
@@ -130,139 +132,318 @@ def solve_analytical(self, torque, psi, i_e):
...
+
+            try:
+                res1 = minimize(obj_torque_err, x0, method='SLSQP',
+                                bounds=bnds, constraints=cons_soft,
+                                options=dict(maxiter=200, ftol=1e-9, disp=False))
+                if res1.success and feasible(*res1.x):
+                    res2 = minimize(obj_loss, res1.x, method='SLSQP',
+                                    bounds=bnds, constraints=cons_soft,
+                                    options=dict(maxiter=200, ftol=1e-9, disp=False))
+                    if res2.success and feasible(*res2.x):
```
max-schenke: When the first minimize is successful, then the second one is run without any consideration of torque, i.e., the starting point (res1.x) may result in the correct torque, but the minimization only tries to reduce losses and doesnt seem to care for providing the referenced torque.

PR #288, opened 2025-10-21
Sphinx conroldoc

bhk11:
Missing Layout changes have to be applied for the gem-control part.


bhk11:
Merged with PR #289 into the nightly.

